PR #21, opened 2018-12-04
Mamba tests

Adrian-St:
Adds 100% test coverage and some abstraction.


application/base.py
```diff
@@ -65,6 +79,11 @@ def collection(cls, index=True, create=True):
     def member(cls, show=True, update=True, delete=True):
         """ Creates the member class with allowed methods """
         methods = {}
+
+        decorators = cls.method_decorators + \
+            list(set(cls.member_decorators) - set(cls.method_decorators))
```
Askir, line 89: what exactly do these two lines do? What does the backslash do? And why do we add the method decorators and then subtract them again? Looks like gibberish to me, maybe a comment would also be nice :D

application/controllers/artifacts_controller.py
```diff
@@ -75,9 +82,6 @@ def show(self, object_id):
     def index(self):
         "Logic for querying several artifacts"
 
-        if not current_app.es:
-            return {"error": "search engine not available"}, 503
```
Askir, line 79: removing all this code duplication is amazing for sure!

application/controllers/artifacts_controller.py
```diff
@@ -139,15 +132,13 @@ def delete(self, object_id):
...
         params = add_tags_params()
         try:
             artifact = Artifact.find(object_id)
-            print(vars(artifact))
-            new_list = getattr(artifact, "tags") + params["tags"]
-            artifact.update({
-                "tags": list(set(new_list))})
+            existing_tags = getattr(artifact, "tags")
+            new_list = existing_tags + list(set(params["tags"]) - set(existing_tags))
```
Askir, line 139: Can't you just call set() again to make everything unique instead of doing `-set(existing_tags)`?

application/models/esmodel.py
```diff
@@ -84,20 +84,17 @@ def save(self):
...
-                body=body)
-            return self, 201
-        except ConflictError:
-            raise NotSaved
+        current_app.es.index(
+            index=self.index,
+            doc_type=self.type,
+            id=self.id,
+            body=body)
+        return self, 201
```
Askir, line 92: Why exactly did you remove the try catch here? There had to be a reason for it before no?

application/controllers/artifacts_controller.py
```diff
@@ -57,16 +57,23 @@ def allowed_file(filename):
...
+        if not current_app.es:
+            return {"error": "search engine not available"}, 503
+        return func(*args, **kwargs)
+
+    return func_wrapper
 
 class ArtifactsController(ApplicationController):
     """ Controller for Artifacts """
 
+    method_decorators = [check_es]
```
Askir, line 72: I don't quite get how this results in check_es being called before the execution of every method. Do you have a link that I could read up on? Is method_decorators a special keyword? Probably not but how exactly does the code in base.py ensure the calling? I'm confused by all this meta programming :D

specs/models/esmodel_spec.py
```diff
@@ -0,0 +1,38 @@
...
+            with Mock() as elastic_mock:
+                elastic_mock.update(ANY_ARG)
+                elastic_mock.index(ANY_ARG)
+
+            current_app.es = elastic_mock
+            self.es_model = ESModel({"type": "image"})
+            self.es_model.index = "artifact"
+
+        with it("raises NotInitialized error if it wasn't saved before"):
+            expect(lambda: self.es_model.update({})).to(raise_error(NotInitialized))
```
Askir, line 34: Testing the ESModel without an ES connection for sure is annoying.

specs/requests/artifacts_spec.py
```diff
@@ -0,0 +1,278 @@
...
+from expects import expect, equal, have_key
+from elasticsearch.exceptions import NotFoundError
+from doublex import Mock, Stub, ANY_ARG
+from specs.spec_helpers import Context
+
+sys.path.insert(0, 'specs')
+
+with description('/images') as self:
+
+    # pylint: disable=duplicate-code
```
Askir: Does it really complain about this? oof, we really need to figure out how to use pylint effectively. There have to be some better presets.

specs/requests/artifacts_spec.py
```diff
@@ -0,0 +1,278 @@
...
+                            {"_index": "artifact",
+                             "_type": "image",
+                             "_id": 1,
+                             "_score": 0.5,
+                             "_source": {
+                                 "created_at": "yesterday",
+                                 "updated_at": "yesterday",
+                                 "file_url": "use_case_diagram.png",
+                                 "tags": ["uml", "use case diagram"],
+                                 "file_date": "yesterday"}}]}})
```
Askir: Can we extract this JSON into a test helper for elastic search responses? Like basically an elastic_search_response_factory. I think they'll be useful later on and reading something like `es_response_factory.list_of_artifacts(2)` is more readable than 20 lines of json.

specs/requests/artifacts_spec.py
```diff
@@ -0,0 +1,278 @@
...
+                        })
+
+                with it('returns a 200 status code'):
+                    expect(self.response.status_code).to(equal(200))
+
+                with it("returns an id"):
+                    expect(self.response.json).to(have_key("id"))
+
+                with it("returns a file_url"):
+                    expect(self.response.json).to(have_key("file_url"))
```
Askir, line 65: If we do this thourougly our tests could just be our API documentation. We just have to link the files in the correct places. These would at least always be up to date, although they don't quite contain the amount of example that I would wish for in a documentation.

specs/requests/artifacts_spec.py
```diff
@@ -0,0 +1,278 @@
...
+
+            with context("without file attached"):
+                with before.each:
+                    self.response = self.context.client().post(
+                        "/images", content_type='multipart/form-data', data={
+                            "tags": []
+                        })
+
+                with it('returns a 400 status code'):
+                    expect(self.response.status_code).to(equal(400))
```
Askir, line 75: why is this not a 422?

specs/requests/artifacts_spec.py
```diff
@@ -0,0 +1,278 @@
...
+
+            with context("with malicious file attached"):
+                with before.each:
+                    self.response = self.context.client().post(
+                        "/images", content_type='multipart/form-data', data={
+                            "image": (BytesIO(b'oof'), 'malicious_file.exe')
+                        })
+
+                with it('returns a 400 status code'):
+                    expect(self.response.status_code).to(equal(400))
```
Askir: Again a 400 that is rather unspecific.

specs/requests/artifacts_spec.py
```diff
@@ -0,0 +1,278 @@
...
+
+    with description('/1'):
```
Askir: can we rename this to '/:id' ?

specs/requests/artifacts_spec.py
```diff
@@ -0,0 +1,278 @@
...
+                        elastic_mock.get(
+                            doc_type='_all', id='1',
+                            index='artifact').returns(
+                                {"_id": "1",
+                                 "_type": "image",
+                                 "_source": {"tags": ["class_diagram", ""],
+                                             "created_at": "today",
+                                             "updated_at": "today",
+                                             "file_url": "test.png",
+                                             "file_date": "today"}})
```
Askir: Another elastic search response for our factory ;)

specs/requests/artifacts_spec.py
```diff
@@ -0,0 +1,278 @@
...
```
Askir: More jsons for our factory <3

specs/requests/artifacts_spec.py
```diff
@@ -0,0 +1,278 @@
...
+
+                        elastic_mock.update(
+                            doc_type='image', id='1', index='artifact',
+                            body={'doc': {"tags": ["class_diagram", "", "new_tag"]}})
+
+                    current_app.es = elastic_mock
+                    self.response = self.context.client().post("/images/1/tags",
+                                                               json={"tags": [
+                                                                   "new_tag", "class_diagram"
+                                                               ]})
```
Askir, line 206: You actually test if it appends, nice!

specs/requests/artifacts_spec.py
```diff
@@ -0,0 +1,278 @@
...
+                                 "updated_at": "yesterday",
+                                 "file_url": "use_case_diagram.png",
+                                 "tags": ["uml", "use case diagram"],
+                                 "file_date": "yesterday"}}]}})
+
+                current_app.es = elastic_mock
+                self.response = self.context.client().get("/images")
+
+            with it('returns a 200 status code'):
+                expect(self.response.status_code).to(equal(200))
```
Askir, line 42: This doesn'T check for the response format and also doesn't check if offset, limit and search term have been passed correctly.

Askir (changes requested):
Most importantly I'd like to have the elasticsearc response factory for testing. And an explanation as why the decorator thing actually works. Everything else is just additional and nice to have.

application/base.py
Adrian-St: this adds the list of member_decorators(decorators that are only called on all members) to the list of all method decorators. The `/` is there to break the line into two lines. I will add a comment :D

Adrian-St: and we use the subtraction of sets to eliminate duplicates

application/controllers/artifacts_controller.py
Adrian-St: Unfortunately just calling `list(set())` somehow gives you a non-determenistic ordering (don't ask me why :D) which is hard to test. Also it kind of makes sense to always have the existing tags first.

application/models/esmodel.py
Adrian-St: before I used `es.create`, which gives you a conflict error if the object with that id already exists. Now we use `index` in the save method, which will replace an existing one.

application/controllers/artifacts_controller.py
Adrian-St: you can define method_decorators on resources with flask-restful [flask-restful](https://flask-restful.readthedocs.io/en/latest/extending.html#resource-method-decorators). I'm just adding that to the Resource in base.

specs/requests/artifacts_spec.py
Adrian-St: It actually doesn't anymore, I just removed the __init__.py, so they are not in the same package anymore. We have to find a solution if we need packages though.

specs/requests/artifacts_spec.py
Adrian-St: Ask flask.restful :D

specs/requests/artifacts_spec.py
Adrian-St: I can return a 422 here

Askir (approved):
Thanks for the changes
, looks much better. I have checked on the input validation. Apparently reqparser is deprecated instead we should use marshmallow. I've already started a branch that changes this which will fix the 400 422 error code thing. Tests are still very long wonder if we should split it into more files? 300 lines aint no joke.  
Anyways looks good to me for now!

PR #22, opened 2018-12-09
Adds input validation erros with web_args/marshmallow

Askir:
Input validation with reqparser is deprecated so I swapped it to web_args which uses marshmallow under the hood.

Also we know have 422 errorcodes when the input is not valid and passed ids have to be uuids :)

### DIFF
--- a/specs/requests/artifacts_spec.py
+++ b/specs/requests/artifacts_spec.py
@@ -9,6 +9,7 @@ from elasticsearch.exceptions import NotFoundError
 from doublex import Mock, Stub, ANY_ARG
 from specs.spec_helpers import Context
 from specs.factories.elasticsearch import es_search_response, es_get_response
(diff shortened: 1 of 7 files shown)


specs/requests/artifacts_spec.py
```diff
@@ -3,29 +3,34 @@
...
+
 with description('/images') as self:
 
     with before.each:
         self.context = Context()
         current_app.es = Stub()
 
-    with after.each:
-        self.context.delete()
+    # with after.each:
```
henleo: halo i bims uncomment me please

specs/requests/artifacts_spec.py
```diff
@@ -3,29 +3,34 @@
...
+        # self.context.delete()
```
henleo: #metoo

PR #30, opened 2018-12-19
Add inital rough Team model with DB connection.

Askir:


.env.example
```diff
@@ -1,4 +1,5 @@
 ES_URL=http://localhost:9200
+NEO4J_URL="0.0.0.0"
```
erksch: Why `" "`?

Pipfile
```diff
@@ -13,6 +13,7 @@ flask-cors = "*"
 webargs = "*"
 flask-socketio = "*"
 eventlet = "*"
+py2neo = "*"
```
erksch, line 16: py2? Is that alright? :D

README.md
```diff
@@ -12,10 +12,25 @@
   - In project folder run `pipenv install`
 
 - Copy .env.example into new .env file
+- It's highly recommended you have docker installed it makes environment setup much easier: https://www.docker.com/get-started
+- Install Neo4j
+  - Neo4J is our GraphDB you can find more info on it here: https://neo4j.com/
+  - To simply run the application locally setup a docker container with neo4j with this command: 
+    ```
+    docker run \
```
erksch, line 21: Is this command Windows friendly?

application/models/esmodel.py
```diff
@@ -7,7 +7,7 @@
 from application.errors import NotFound, NotInitialized
 
 
-class ESModel():
+class ESModel:
```
erksch, line 10: Why? Good chance for me to learn more python syntax :D

application/models/team.py
```diff
@@ -0,0 +1,63 @@
+from application.errors import NotFound, NotInitialized
+from flask import current_app
+from py2neo import Graph, Node
+import uuid
+
+
+class Team:
+
+    @classmethod
+    def all(cls):
```
erksch, line 12: What is `cls`?

application/models/team.py
```diff
@@ -0,0 +1,63 @@
...
+        results = current_app.graph.run("MATCH (x:Team) RETURN x")
+        teams = []
+        for result in results:
+            team = Team(result['x']['name'])
```
erksch: This looks weird to me... why do you use `Team` within the same class?
I mean I understand, but wouldn't it be better to separate the `Team` model and the queries?

application/models/team.py
```diff
@@ -0,0 +1,63 @@
...
+            team.id = result['x']['id']
+            teams.append(team)
+        return teams
+
+    @classmethod
+    def find_by(cls, force=False, id='', name=''):
+        team_node = current_app.graph.run(
+            f"MATCH (x:Team) WHERE x.name = '{name}' OR x.id = '{id}' RETURN x").evaluate()
```
erksch: Is the code styling correct?
```
current_app.graph.run(
    ...).evaluate()
```
Looks pretty weird to me.

application/models/team.py
```diff
@@ -0,0 +1,63 @@
...
+        results = current_app.graph.run("MATCH (x:Team) RETURN x")
+        teams = []
+        for result in results:
+            team = Team(result['x']['name'])
+            team.id = result['x']['id']
+            teams.append(team)
+        return teams
+
+    @classmethod
+    def find_by(cls, force=False, id='', name=''):
```
erksch: Hardcoding the search parameters in the arguments does not scale very well.
Maybe you can pass a hash of search parameters and append an `OR x.{key} = '{value}'` to the query for each member of the hash.

erksch (changes requested):
Great work Jascha! Very comprehensible Team class! I'm hyped to use it :+1: 
Just a few questions and notes.

.env.example
Askir: Had trouble with the IP earlier but I can try removing them and see if it still works.

Pipfile
Askir: py2neo is the python library for neo4j doesn't have anything to do with python 2 or python 3

README.md
Askir: Oh good question, how do you setup docker containers on windows? It's probably not, just copy pasted from the neo4j website

application/models/esmodel.py
Askir: Braces are for inheritance, if you don't inherit there is no need for braces. At least that's what PyCharm suggested.

application/models/team.py
Askir: All class methods in python do have cls (class) as the first parameter that is always implicitly passed to the method.
normal instance methods have self passed instead.

application/models/team.py
Askir: I'm going for an ActiveRecord like model. So you can say.: `Team.all()` and it will return all the Team Objects as an Array. Yes you could extract this to a different class but seems like upfront design to do that for just one model.
Abstracting Elasticsearch and the Graphdatabase in a useful way will be a very interesting task nontheless.

application/models/team.py
Askir: Python doesn't do very well with method calls over multiple lines especially when the argument is only one string. I'll extract that String into a variable and pass that to the function instead, should make it more readible.

application/models/team.py
Askir: You have to adjust the Cypher query individually for all search params anyways, and I don't think we will add THAT many fields to the Team Node. I agree that a hash would maybe be nice but then the function call also looks somewhat messy with `Team.find_by({"id": some_uuid})` instead of `Team.find_by(id=some_uuid)`

I'll have a look at how django does it maybe.

application/models/team.py
Adrian-St: If you put expressions in brackets, you can split them into multiple lines.
For strings you can use:
```
 ("first sentence"
  "second sentence")
```
which results in: "first sentencesecond sentence"

application/models/team.py
Adrian-St: I like the params hash more too, alternatively you could use `find_by(**kwargs)` for passing an arbitrary number of named parameters.

application/models/team.py
```diff
@@ -0,0 +1,78 @@
...
+        if force:
+            raise NotFound()
+        else:
+            return None
+
+    @classmethod
+    def _find_team_node(cls, id_='', name=''):
+        query = f"MATCH (x:Team) WHERE x.name = '{name}' OR x.id = '{id_}' RETURN x"
+        return current_app.graph.run(query).evaluate()
+
```
Adrian-St, line 44: If you don't need the cls argument for helper methods it's better
to use `@staticmethod` where you don't pass `cls` as first parameter.

wait_for_it.sh
```diff
@@ -0,0 +1,178 @@
...
+
+if [[ $WAITFORIT_CLI != "" ]]; then
+    if [[ $WAITFORIT_RESULT -ne 0 && $WAITFORIT_STRICT -eq 1 ]]; then
+        echoerr "$WAITFORIT_cmdname: strict mode, refusing to execute subprocess"
+        exit $WAITFORIT_RESULT
+    fi
+    exec "${WAITFORIT_CLI[@]}"
+else
+    exit $WAITFORIT_RESULT
+fi
```
Adrian-St, line 178: Did you write that yourself?
If not, please put a link in the comments so we know where we get third party code from.

Adrian-St (changes requested):
Looks really good! 
Just use @staticmethod where you don't need the class
and then we can merge.

application/models/team.py
Askir: This is outdated and splitting into multiple lines with braces is pretty much unreadable imo. I try to avoid it wherever possible.

application/models/team.py
Askir: yes **kwargs should be used for this but not right now as there are only 2 params to use. Will be refactored once we add more to the model.

application/models/team.py
Askir: If the method is static you can call it from an instance instead of on a class level. Which in this case wouldn't make any sense. I like containing the scope of this function this way.

wait_for_it.sh
Askir: No I did not it is the first result if you google wait for it sh though. Simple script that helps with docker-compose. I could load it via git each time but I think that would be make CI even slower :D

application/models/team.py
Askir: Conceptually a static method is a method that knows nothing about the class it could just as well exist without it's calls. You could even write it outside of the class scope. But this method doesn't make any sense at all without the team class which is why I think class method is more fitting personally.

wait_for_it.sh
Askir: This doesn't have anything to do with the application, consider it a bandaid for now.

PR #33, opened 2019-01-07
Team routes

henleo:


Askir (commented):
Looks pretty clean so far. I'll add the additional team route changes to this branch as well so if you force push please notify me @henleo @arne-z

Askir:
Also an update route is still missing, is that not intended to exist? Changing a teams name might be important later :)

application/controllers/teams_controller.py
```diff
@@ -0,0 +1,31 @@
...
+from application.base import respond_with
+from .application_controller import ApplicationController
+
+
+class TeamsController(ApplicationController):
+    """ Controller for teams """
+
+    def show(self, object_id):
+        try:
+            return respond_with(Team.find_by(force=True, id_=object_id)), 200
```
erksch: A little too much going on in one line in my opinion.
Why not 
```
team = Team.find_by(...)
return respond_with(team), 200
```

application/schema.py
```diff
@@ -24,3 +25,16 @@ def build_url(self, file_url):
         """ Schema: fileserver/id_filename """
         return current_app.config["FILE_SERVER"] + \
                 "/" + file_url
+
+
+class TeamSchema(BaseSchema):
```
erksch: Why multiple classes in one file? Maybe extract schemas into separate files?

application/validators/teams_validator.py
```diff
@@ -0,0 +1,22 @@
+""" Defines validators for team requests """
+
+from webargs import fields, ValidationError
+
+
+def get_args():
```
erksch: Why is it called `get_args` instead of `index_args`?

erksch (changes requested):

erksch:
Please add an issue number to the branch name next time!

specs/requests/teams_spec.py
```diff
@@ -0,0 +1,95 @@
...
+        with description('GET'):
+            with description('valid id'):
+                with before.each:
+                    self.team = Team.create(name='Blue')
+                    self.response = self.context.client().get(f"/teams/{self.team.id_}")
+
+                with it('returns 200 ok'):
+                    expect(self.response.status_code).to(equal(200))
+
+                with it('returns the correct image'):
```
erksch: Copy paste busted hehehe!

specs/requests/teams_spec.py
```diff
@@ -0,0 +1,95 @@
...
+    with description('POST'):
+        with description('valid request'):
+            with before.each:
+                self.response = self.context.client().post(
+                    "/teams",
+                    content_type='multipart/form-data',
+                    data={
+                        "name": "My Team"
+                    })
+            with it('responds with 200'):
```
erksch, line 45: The missing newline here is inconsistent :(((

erksch (changes requested):

specs/requests/teams_spec.py
```diff
@@ -0,0 +1,95 @@
...
+                have_key("name", "Blue"),
+                have_key("name", "Red")
+            ))
+
+    with description('POST'):
+        with description('valid request'):
+            with before.each:
+                self.response = self.context.client().post(
+                    "/teams",
+                    content_type='multipart/form-data',
```
erksch: Copy paste detected lel! Should be `application/json`

erksch (changes requested):

specs/requests/teams_spec.py
```diff
@@ -0,0 +1,95 @@
...
+                expect(Team.exists(id_=self.response.json["id"], name="My Name")).to(be(True))
+
+        with description('invalid request'):
+            with before.each:
+                self.response = self.context.client().post(
+                    "/teams",
+                    content_type='application/json',
+                    data={
+                        "name": ""
+                    })
```
erksch: Can somebody explain to me when to use which quotation marks in python?

specs/requests/teams_spec.py
```diff
@@ -0,0 +1,95 @@
...
+                with before.each:
+                    self.team = Team.create(name='Blue')
+                    self.response = self.context.client().get(f"/teams/{self.team.id_}")
+
+                with it('returns 200 ok'):
+                    expect(self.response.status_code).to(equal(200))
+
+                with it('returns the correct image'):
+                    expect(self.response.json).to(have_key("name", "Blue"))
+                    expect(self.response.json).to(have_key("id", self.team.id_.urn[9:]))
```
erksch, line 79: Dafuq does `.urn[9:]` mean?

specs/requests/teams_spec.py
```diff
@@ -0,0 +1,95 @@
...
+            with description('valid id'):
+                with before.each:
+                    self.team = Team.create(name='Blue')
+                    self.response = self.context.client().get(f"/teams/{self.team.id_}")
+
+                with it('returns 200 ok'):
+                    expect(self.response.status_code).to(equal(200))
+
+                with it('returns the correct image'):
+                    expect(self.response.json).to(have_key("name", "Blue"))
```
erksch, line 78: Why not use `self.team.name` instead of `"Blue"`? Idk maybe I'm wrong @Askir?

specs/requests/teams_spec.py
```diff
@@ -0,0 +1,95 @@
...
+                    expect(self.response.json).to(have_key("id", self.team.id_.urn[9:]))
+
+            with description('invalid id'):
+                with before.each:
+                    team_blue = Team.create(name='Blue')
```
erksch: I don't think you need team_blue here.

erksch (changes requested):

specs/requests/teams_spec.py
erksch: Also by this, remove the naming `team_orange` afterwards.

specs/requests/teams_spec.py
erksch: Do you even need to set the `content_type`?

erksch: Okay, Arne and I figured out you don't need to set any `content_type` on these type of requests furthermore it actually broke some tests to use `content_type` in the first place. So I advise to just leave it out unless you really need it somehow.

specs/requests/teams_spec.py
erksch: Okay Arne explained it to me.

erksch (approved):

application/validators/teams_validator.py
```diff
@@ -0,0 +1,32 @@
...
+    return {"name": fields.Function(
+        required=True,
+        deserialize=validate_team_name
+    )}
+
+def update_args():
+    """Defines and validates params for update"""
+    return {
+        "id": fields.UUID(required=True, load_from='object_id', location='view_args'),
+        "name": fields.Function(deserialize=validate_team_name)
```
Askir: this is rewriting inbuild webargs functionality, just do `fields.String(required=True)` The error messages will be autogenerated

specs/requests/teams_spec.py
```diff
@@ -0,0 +1,129 @@
...
+                    self.team = Team.create(name='Blue')
+                    self.response = self.context.client().put(
+                        f"/teams/{self.team.id_}",
+                        data={"name": "Red"})
+
+                with it('responds with 200'):
+                    expect(self.response.status_code).to(equal(200))
+
+                with it('updates the team correctly'):
+                    fresh_team = Team.find_by(id_=self.team.id_)
```
Askir: this line should be in the before block. Or if you don't want to execute it every time because it is just needed once create another description block: one for the response and one fore the updated team. Because you aren't using the `self.response` in this statement either.

specs/requests/teams_spec.py
```diff
@@ -0,0 +1,129 @@
...
+                    expect(fresh_team.name).to(equal("Red"))
+
+                with it('responds with the updated team'):
+                    expect(self.response.json).to(have_key("name", "Red"))
+                    expect(self.response.json).to(have_key("id", self.team.id_.urn[9:]))
+
+            with description('invalid id'):
+                with before.each:
+                    team = Team.create(name="Red")
+                    team.delete()
```
Askir, line 114: Any special reason to create and immediatly delete a team again?

specs/requests/teams_spec.py
```diff
@@ -0,0 +1,129 @@
...
+from specs.factories.elasticsearch import es_search_response, es_get_response
+from specs.factories.uuid_fixture import get_uuid
+from specs.factories.date_fixture import get_date, date_regex
+
+sys.path.insert(0, 'specs')
+
+with description('/teams') as self:
+    with before.each:
+        self.context = Context()
+        current_app.es = Stub()
```
Askir: Pretty sure this line isn't needed in this test file

Askir (changes requested):

Askir (approved):

PR #38, opened 2019-01-11
9/image recognition

arne-z:
Implements image recognition using the Google cloud vision API.


Askir:
I had to reroll our dependency changes because it bricked some stuff on CI since it would update a lot of our dependecies to prereleases and major version releases which simply bricked about half our tests.
We should just do that dependency update in a seperate branch and also consider using something different than pipenv: https://chriswarrick.com/blog/2018/07/17/pipenv-promises-a-lot-delivers-very-little/
I personally think poetry looks a little bit cleaner but I'd appreciate some input: https://github.com/sdispater/poetry

application/controllers/artifacts_controller.py
```diff
@@ -38,30 +41,31 @@ def show(self, object_id):
...
         filename = str(uuid.uuid4()) + "_" + \
-            werkzeug.utils.secure_filename(uploaded_file.filename)
+                   werkzeug.utils.secure_filename(uploaded_file.filename)
         file_url = os.path.join(current_app.config["UPLOAD_FOLDER"], filename)
         uploaded_file.save(file_url)
         params["file_url"] = filename
         artifact = Artifact(params)
 
         artifact.save()
+        recognize_image(file_url, artifact)
```
Askir: This method is getting pretty long. Tbh it was way too long before already.  
Simply doing some method extraction would help imo. So something like this?
```suggestion
        params["file_url"] = upload_file()
        artifact = Artifact(params)
        artifact.save()
        recognize_image(params["file_url", artifact)
```

also consider renaming `recognize_image()` to something like 'add_google_vision_tags()'

application/helpers/artifact_helper.py
```diff
@@ -0,0 +1,8 @@
+"""Defines helper methods used with artifacts"""
+from flask import current_app
+
+
+def build_url(file_url):
+    """ Schema: fileserver/id_filename """
+    return current_app.config["FILE_SERVER"] + \
+        "/" + file_url
```
Askir: Really not a fan of the name "helper" that could basically be anything and nobody knows where to use it and what to put here.

I think the reason you extracted it from the schema is because you wanted to use the method outside the Schemaclass. But it never was a private method to begin with. And it doesn't need to be an instance method either. If you just change it to a class Method you can call it via: `ArtifactSchema.build_url()` which I think is sufficient and we don't need an extra folder + module for one method.

application/recognition/image_recognition.py
```diff
@@ -0,0 +1,48 @@
...
+def recognize_image(file_url, object_id):
+    """Called by classes """
+    spawn_n(_work_asynchronously, file_url, object_id)
+
+
+def _call_google_api(url):
+    """Does API call to could vision api"""
+    api_url = current_app.config.get('CLOUD_VISION_API_URL')
+    api_key = current_app.config.get('CLOUD_VISION_API_KEY')
+    r = post(url=api_url,
```
Askir: What if the image api is not available? Does anything significant happen? If it just crashes it's probably not too bad just want to prevent that image upload fails if the API is unavailable. But since it is in a second thread this probably doesn't brick that right?

application/recognition/image_recognition.py
```diff
@@ -0,0 +1,48 @@
...
+                     }
+                 ]
+             },
+             params={'key': api_key})
+    return r.json()
+
+
+def _work_asynchronously(file_url, artifact):
+    """Private method called asynchronously for image recognition."""
+    image_url = build_url(file_url)
```
Askir: you could also probably just add a method to artifact: `artifact.current_url()` or something similar.

application/recognition/image_recognition.py
```diff
@@ -0,0 +1,48 @@
...
+    labels = res['responses'][0]['labelAnnotations']
+    new_tags = []
+    for label in labels:
+        new_tags.append(label['description'])
+    texts = res['responses'][0]['textAnnotations']
+    for text in texts:
+        new_tags.append(text['description'])
+    existing_tags = artifact.tags or []
+    tags = existing_tags + new_tags
+    artifact.update({"tags": tags})
```
Askir: Again some method extraction would help with readability here:
```suggestion
    res = _call_google_api()
    new_tags = extract_tags(res)
    update_tags(artifact, new_tags)
```

application/schemas/artifact_schema.py
```diff
@@ -1,11 +1,12 @@
-""" Defines schema for database Artifact objects """
+""" Defines schema for database artifact objects """
```
Askir, line 1: I mean the class is literally named Artifact. I think keeping it with a capital makes a lot of sense.

application/schemas/artifact_schema.py
```diff
@@ -19,10 +20,5 @@ class ArtifactSchema(BaseSchema):
...
     def transform_fields(self, data):
         """ Transforms field for output """
-        data["url"] = self.build_url(data.pop("file_url"))
+        data["url"] = build_url(data.pop("file_url"))
         return data
-
-    def build_url(self, file_url):
-        """ Schema: fileserver/id_filename """
-        return current_app.config["FILE_SERVER"] + \
-                "/" + file_url
```
Askir, line 28: This I would just mark as a class method and then it can stay in here.

specs/recognition/image_recognition_spec.py
```diff
@@ -0,0 +1,21 @@
...
+
+with description('image recognition called during create') as self:
+    with before.each:
+        self.context = Context()
+        with Mock() as artifact_mock:
+            artifact_mock.update(ANY_ARG).returns(no_content())
+            artifact_mock.tags = []
+        self.artifact = artifact_mock
+    with it('Adds Tags to artifact'):
+        with patch('application.recognition.image_recognition._call_google_api',
```
Askir: you can use the extracted `mock_image_recognition` here as well right?

Askir (changes requested):
Very cool that it works this easily. Only requesting some code style changes :)
Poke me once you need another look at it. Probably wont take very long to fix :D

application/controllers/artifacts_controller.py
arne-z: I've not worked with the artifacts_controller up until now so I will look into where to put the upload_file function, and what has to go into it. I think extracting that as a method sounds reasonable. 
On the rename: 
I would like to keep the rest of ejija agnostic as to what type of recognition is actually used.
This allows us to only change what (private) function the worker thread in recognize_image calls between the local and the google api image recognition. 
This will enable A/B testing between the 2 versions and using them in parrallel, by only adjusting one line in the recognize_image function.

application/recognition/image_recognition.py
arne-z: Since the worker thread is created with the spawn_n method, any error, exception or return value will never get back to the main thread. The worker thread is forgotten the moment it is created.

The thread itself can fail at this point, for multiple reasons (e.g. incorrect .env file or a 4xx answer from the API), but this will not get back to the main thread.

application/recognition/image_recognition.py
arne-z: Agreed.

application/schemas/artifact_schema.py
arne-z: just consistency with the other files arounf it. They are usually non-capital artifacts in the docstring (see controller and model).
I really don't feel strongly about this so I'm fine with "Artifacts".

specs/recognition/image_recognition_spec.py
arne-z: no, this test is intended to test the image_recognition function, so mocking it would achieve nothing.
Instead the _call_google_api function is mocked, so I don't actually do an http call, and instead give back a canned response from the google api, as well as the update function of the artifact, so I don't actually update the database.

I thought I read, that the Mock will raise an error if it is not called during test execution, but if this is not the case, I need to find a better mock.

application/controllers/artifacts_controller.py
Askir: If you don't want to specify google for that reason, that's okay but on first sight `recognize_image` doesn't really tell me what it does. For me that sounds more like, the function will check weather the system already knows the image or something like that and not automatically add tags. So if you want to rename to `auto_add_tags` that's fine with me as well.

application/recognition/image_recognition.py
Askir: Alright then!

application/schemas/artifact_schema.py
Askir: hmm sure I didn't see the other files in this PR but that makes sense. Consistency is more important to me as well although I would prefer the the capital.

specs/recognition/image_recognition_spec.py
Askir: Oh I see that makes more sense. In general I am not a big fan of the patch call. Simply because it doesn't read very well and you have to change it if you move the file. Any specific reason for not making the image_recognition module a class? Then you could simply do:
```
image_recognition = ImageRecognition
Mock(image_recognition._call_google_api).returns(vision_api_response)
image_recognition._work_asynchonously()
```
I think that would be more clear and with doublex-expects, you could also do an expect statement like:
`expect(image_recognition._call_google_api()).to(have_been_called.once)`

It's already installed, documentation can be found here: https://github.com/jaimegildesagredo/doublex-expects

specs/recognition/image_recognition_spec.py
```diff
@@ -0,0 +1,21 @@
...
+                   return_value=vision_api_response()):
+            expect(_work_asynchronously('files.askir.me/test.jpg', self.artifact)).to(be_none)
```
Askir: Just a note here btw, in general testing private methods is considered to be bad style. You should only test the behaviour of the method not the implementation. But since you are only doing this to test the asynchronous nature, this is a case where testing a private method is necessary.

I'm not sure if you knew this so I just wanted to mention it.

application/controllers/artifacts_controller.py
arne-z: Sounds fine.

application/helpers/artifact_helper.py
arne-z: This is now a static method in artifact schema.

application/recognition/image_recognition.py
arne-z: This is now a static method in artifact schema.

application/schemas/artifact_schema.py
arne-z: This is now a static method in artifact schema.

specs/recognition/image_recognition_spec.py
arne-z: This is really hard, spent some hours and I'm not sure it is possible.
It works for artifacts_controller_spec because I don't want the class to do anything, but it fails to work for image_recognition_spec because once I mock the image recognition class, there is no way to make it actually execute any function.
Obviously, I can't test a class, by mocking the class itself.

arne-z: This can be used in some places but does not get rid of the Patch entirely.
I was able to mock the method in image_recognition_spec but in articaft_controller_spec I still need to use patch, since the class is not actually available in scope for mocking.

specs/recognition/image_recognition_spec.py
arne-z: I tried out testing the auto_add_tags function, but it appears that mocks do not carry over onto the worker thread, or at the very least never hear back from the worker thread, so running the tests including the thread spawn does not help.

application/recognition/image_recognition.py
```diff
@@ -0,0 +1,56 @@
...
+                             "image": {
+                                 "source": {"imageUri": url}
+                             }
+                         }
+                     ]
+                 },
+                 params={'key': api_key})
+        return r.json()
+
+    @staticmethod
```
Askir, line 36: Not a huge fan of having a class with only static methods. I'd prefer most of these to simply be class methods so if you want to use class variables or later on intialize objects you easily can. The difference is miniscule but for example a static private function doesn't make much sense. Since the static function doesn't even belong to a class how is it supposed to be private? A private function should not be called from outside the class but the static method doesn't even belong to a class so every call is from outside...  
Just feels not right to me at least.

application/schemas/artifact_schema.py
Askir: Again I'd prefer a class method here, althought this one is supposed to be public so it doesn't really matter. Just my preference.

specs/recognition/image_recognition_spec.py
```diff
@@ -0,0 +1,24 @@
...
+        with Mock() as artifact_mock:
+            artifact_mock.update(ANY_ARG).returns(no_content())
+            artifact_mock.tags = []
+            artifact_mock.file_url = 'files.askir.me/test.jpg'
+        self.artifact = artifact_mock
+        self.recognizer._call_google_api = method_returning(vision_api_response())
+    with it('Adds Tags to artifact'):
+        self.recognizer._work_asynchronously(self.artifact)
+        expect(self.recognizer._call_google_api).to(have_been_called.once)
+        expect(self.artifact.update).to(have_been_called.once)
```
Askir, line 24: This looks pretty clean now. gj

Askir (approved):
Other than my preference of class over static methods this looks pretty clean now. So go ahead and merge once you are happy yourself!

PR #40, opened 2019-01-12
Update pyyaml

Askir:
Update pyyaml dependency.


Askir:
meh, once you add one --pre dependency pipenv keeps updating every pre dependency which ends up with marshmallow being updated to 3.x from 2.x, which breaks our current implementation. Bah! Welp I'll fix this some other time, it's only coveralls that is affected by the security vulnerability anyways so at worst people can push other results to our coveralls... if they get the tokens. I mean I'm scared of 0% but not THAT scared!

erksch:
Closing. Please try again with the new peotry setup.

PR #41, opened 2019-01-16
Team presentation view

Adrian-St:
Adds new route for requests coming from dialogflow, that pushes 
images directly to the frotend via websockets when called.
Also add team_name to the parameters.


Askir (commented):
haven't looked at everything yet but I already like how you changed the imports to relative paths :+1:

application/controllers/dialogflow_artifacts_controller.py
```diff
@@ -0,0 +1,31 @@
+"""
+Handles all logic of the dialogflow_artefacts api
```
erksch: *dialogflow_art**i**facts

docker-compose.yml
```diff
@@ -20,14 +20,14 @@ services:
...
      - "7474:7474"
-  elasticsearch_dev:
-    image: docker.elastic.co/elasticsearch/elasticsearch:6.5.4
-    environment:
-      - discovery.type=single-node
-      - ES_JAVA_OPTS=-Xms512m -Xmx512m
-    ports:
-      - "9300:9300"
-      - "9200:9200"
+#  elasticsearch_dev:
```
erksch: Please remove the comment, we need this for the correct setup.
If you don't want to start the elasticsearch_dev container just do something like
```
docker-compose up -d
docker-compose stop elasticsearch_dev
``` 
instead of changing the `docker-compose.yml`

specs/factories/request_generatory.py
```diff
@@ -0,0 +1,19 @@
+"""
```
erksch: Does the word generatory exist? :D

erksch (changes requested):

erksch:
I have a further question: We do now have a seperate controller for image requests by dialogflow. Is there any way dialogflow could use the normal /images route and the normal controller? Or can you explain why this would not work?

Adrian-St:
@erksch Requests coming from dialogflow only know the team_name and not the team_id.
Also it sends the START_PRESENTATION to all clients. I don't know how clean it is to merge that together in one route.

erksch (approved):

PR #43, opened 2019-01-16
Clean up wait for it in docker-compose setup

erksch:
I worked on timeout problems with the cypress CI and tried to do some improvements here.
But I kind of ended up just making it prettier than actually improving anything.

Most importantly I replaced [wait-for-it.sh](https://github.com/vishnubob/wait-for-it) by [wait-for.sh](https://github.com/Eficode/wait-for) because wait-for-it.sh is a bash script and wait-for.sh is a shell script and thus compatible with alpine linux images which this project uses to build its docker image. 
This enables us to cut out the `apk add bash &&` and consequentially the `sh -c`. 

I increased the timeout to 30 seconds hoping this will fix problems with the cypress ci.

I also extracted the gunicorn command into a separate script to share between the dockerfile and docker-compose overwrite.


Askir:
Thanks for the detailed write up and the extra work.  
Still not sure if wait for is the way to go. For the extra docker use cases there is also dockerize which can do more and just looks more like an all around good piece of software: https://github.com/jwilder/dockerize

Askir (approved):

PR #47, opened 2019-01-21
Files nginx for local docker-compose 

erksch:
This PR adds an nginx server to the docker-compose setup which serves the files in the uploads folder of the Elija container. This enables to serve the files smoothly when developing locally on the frontend.

### DIFF
--- a/.env.dc
+++ b/.env.dc
@@ -1,6 +1,6 @@
 ES_URL=http://elasticsearch_dev:9200
 NEO4J_URL=bolt://neo4j_dev:7687
-FILE_SERVER=localhost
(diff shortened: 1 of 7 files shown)


.env.dc
```diff
@@ -1,6 +1,6 @@
 ES_URL=http://elasticsearch_dev:9200
 NEO4J_URL=bolt://neo4j_dev:7687
-FILE_SERVER=localhost
+FILE_SERVER=http://localhost:8080
```
erksch, line 3: Since Jona is not included in the dc setup and you would run it outside the docker stuff it's important to set a URL that is accessible from outside.

PR #49, opened 2019-01-21
Multi Tagging

erksch:
This PR enables support for multi-tagging.
Which means: I can set tags for a chunk of images.

My suggested route is `POST /tags` with `{ "ids": [...], "tags": [...] }` 
Please feel free to suggest a better route.

Regarding the algorithm it works more like a PUT:
The set of tags you send over this route will consequentially be the set of tags all the images have in common.

So if you have two images tagged with `[a, b, c]` and `[a, b, d]` and I send `[a]` as the new multi tags for those images the resulting set of tags will be `[a, c]` and `[a, d]` since I stated `[a]` as the **total set of tags the images should have in common**.

Another approach would be to make separate delete and create routes for single multi tags. Maybe worth a thought, too.

### DIFF
--- a/application/controllers/tags_controller.py
+++ b/application/controllers/tags_controller.py
@@ -13,6 +13,32 @@ class TagsController(ApplicationController):
 
     method_decorators = [check_es_connection]
 
+    @use_args(tags_validator.multi_tagging_args())
+    def multi_tagging(self, params):
+        """ Add tags to multiple artifacts """
+
+        new_multi_tags = list(params["tags"])
+        existing_multi_tags = []
+        artifacts = []
+
+        for object_id in set(params["ids"]):
+            try:
+                artifact = Artifact.find(object_id)
+                artifacts.append(artifact)
+
+                if not existing_multi_tags:
+                    existing_multi_tags = artifact.tags
+                else:
+                    existing_multi_tags = list(set(existing_multi_tags) & set(artifact.tags))
+            except NotFound:
+                return {"error": f"{object_id} not found"}, 404
+
+        for artifact in artifacts:
+            artifact.tags = list(set(artifact.tags) - set(existing_multi_tags)) + new_multi_tags
+            artifact.save()
+
+        return '', 204
+
     @use_args(tags_validator.add_tags_args())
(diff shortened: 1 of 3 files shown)


erksch:
I'm sorry tests are on the way :/

erksch:
How about `PUT /tags/multi` or `PUT /multiTags`

application/controllers/tags_controller.py
```diff
@@ -13,6 +13,32 @@ class TagsController(ApplicationController):
 
     method_decorators = [check_es_connection]
 
+    @use_args(tags_validator.multi_tagging_args())
+    def multi_tagging(self, params):
+        """ Add tags to multiple artifacts """
+
+        new_multi_tags = list(params["tags"])
```
Adrian-St, line 20: params["tags"] should be a list already, no need to convert again.

application/controllers/tags_controller.py
```diff
@@ -13,6 +13,32 @@ class TagsController(ApplicationController):
...
+        existing_multi_tags = []
+        artifacts = []
+
+        for object_id in set(params["ids"]):
```
Adrian-St, line 24: Please use Artifact.find_all instead.
Elasticsearch uses mget for multiple id's

application/controllers/tags_controller.py
```diff
@@ -13,6 +13,32 @@ class TagsController(ApplicationController):
...
+            try:
+                artifact = Artifact.find(object_id)
+                artifacts.append(artifact)
+
+                if not existing_multi_tags:
+                    existing_multi_tags = artifact.tags
+                else:
+                    existing_multi_tags = list(set(existing_multi_tags) & set(artifact.tags))
```
Adrian-St, line 32: Why do you always convert to list?
I think it would be better to initialize existiting_multi_tags as an empty set.

Adrian-St (changes requested):

erksch:
Thanks for taking a look, but I discussed my frontend approach with Arne and would favour deprecating this change. Stay tuned for my other approach.

PR #50, opened 2019-01-22
Artifact Update Many

erksch:
Add an update many route on `PATCH /images`

### DIFF
--- a/application/controllers/artifacts_controller.py
+++ b/application/controllers/artifacts_controller.py
@@ -68,6 +68,20 @@ class ArtifactsController(ApplicationController):
         except NotFound:
             return {"error": "not found"}, 404
 
+    @use_args(artifacts_validator.update_many_args())
+    def update_many(self, params):
+        """ Logic for updating multiple artifacts at once """
+
+        for update_data in params["artifacts"]:
+            object_id = update_data.pop("id")
+            try:
+                artifact = Artifact.find(object_id)
+                artifact.update(update_data)
+            except NotFound:
+                return {"error": f"failed at <{object_id}>: not found"}, 404
+
+        return no_content()
+
     @use_args(artifacts_validator.delete_args())
     def delete(self, params, object_id):
         """Logic for deleting an artifact"""
--- a/specs/requests/artifacts_spec.py
+++ b/specs/requests/artifacts_spec.py
@@ -8,6 +8,7 @@ from expects import expect, equal, have_key
 from hamcrest import matches_regexp
 from elasticsearch.exceptions import NotFoundError
 from doublex import Mock, Stub, ANY_ARG
+from doublex_expects import have_been_satisfied
 from specs.spec_helpers import Context
 from specs.factories.elasticsearch import es_search_response, es_get_response
 from specs.factories.uuid_fixture import get_uuid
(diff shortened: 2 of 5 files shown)


application/controllers/artifacts_controller.py
```diff
@@ -68,6 +68,20 @@ def update(self, params, object_id):
         except NotFound:
             return {"error": "not found"}, 404
 
+    @use_args(artifacts_validator.update_many_args())
+    def update_many(self, params):
+        """ Logic for updating multiple artifacts at once """
+
+        for update_params in params["data"]:
```
Askir: rename they key to `images` and the `update_params` to `image` imo. That would be a lot more descriptive.

application/controllers/artifacts_controller.py
```diff
@@ -68,6 +68,20 @@ def update(self, params, object_id):
...
+            object_id = update_params.pop("id")
+            try:
+                artifact = Artifact.find(object_id)
+                artifact.update(update_params)
```
Askir: This can probably be improved later on to not do a database request for every single image but instead do batch processing. Fine for now though.

specs/requests/artifacts_spec.py
```diff
@@ -133,6 +133,73 @@
                 with it('returns a 422 status code'):
                     expect(self.response.status_code).to(equal(422))
 
+        with description('UPDATE MANY'):
+            with context('valid request'):
+                with context("all the resources exist"):
+                    with before.each:
+                        with Mock() as elastic_mock:
+                            for i in range(2):
```
Askir: you can do: `.get(...).returns(...).times(2)` instead. I think that's a little cleaner.

specs/requests/artifacts_spec.py
```diff
@@ -133,6 +133,73 @@
...
+                                },
+                                {
+                                    "id": get_uuid(1),
+                                    "tags": ["blue", "green"]
+                                }
+                            ]
+                        })
+
+                    with it('returns a 204 status code'):
+                        expect(self.response.status_code).to(equal(204))
```
Askir, line 177: I think this deserves some more specific Mock expects with doublex-expects. https://pypi.org/project/doublex-expects/
Simply adding: `expect(current_app.es).to(have_been_satisifed)` would be sufficient I think.

specs/requests/artifacts_spec.py
```diff
@@ -133,6 +133,73 @@
...
+                    with it('returns a 404 status code'):
+                        expect(self.response.status_code).to(equal(404))
+
+                    with it('includes error message'):
+                        expect(self.response.json).to(have_key("error"))
+
+            with context('invalid request'):
+                with before.each:
+                    self.response = self.context.client().patch("/images", json={
+                        "data": [{}]
```
Askir: what happens if you have a uuid as a key here?

Askir (approved):
Just some renaming of variables and stuff, wont block merging but please fix :p

Looks pretty clean otherwise!

application/controllers/artifacts_controller.py
Adrian-St: Already exists, just use Artifact.find_all

Adrian-St (approved):

application/controllers/artifacts_controller.py
erksch: Hmm but why image and not artifacts?

specs/requests/artifacts_spec.py
erksch: I think the world would be a better place...

PR #51, opened 2019-01-23
Add (not so basic anymore) abstract NeoModel.

Askir:
I've implemented a wrapper surrounding the py2neo ogm to fit our prefered way of abstracting. I'm also in the process of adding neo+elastic sync in this PR. Simply by calling methods on both models for now.

Code Reviews greatly appreciated @erksch @Adrian-St

Also an open TODO is using the schema from marshmallow to auto generate the Property() calls. Not sure if that is possible but would be cool.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -27,6 +27,7 @@ doublex-expects = "*"
 pep8 = "*"
 autopep8 = "*"
 rope = "*"
+PyHamcrest = "*"
 
 [requires]
 python_version = "3.7"
--- a/application/controllers/artifacts_controller.py
+++ b/application/controllers/artifacts_controller.py
@@ -15,6 +15,7 @@ from ..validators import artifacts_validator
 from ..recognition.image_recognition import ImageRecognizer
 from .application_controller import ApplicationController
 
+
 class ArtifactsController(ApplicationController):
     """ Controller for Artifacts """
 
@@ -22,10 +23,9 @@ class ArtifactsController(ApplicationController):
 
     def show(self, object_id):
         """Logic for getting a single artifact"""
-
         try:
             artifact = Artifact.find(object_id)
-            return respond_with(artifact)
+            return respond_with(artifact.elastic)
         except NotFound:
             return {"error": "not found"}, 404
 
@@ -40,12 +40,16 @@ class ArtifactsController(ApplicationController):
     @use_args(artifacts_validator.create_args())
     def create(self, params):
         """Logic for creating an artifact"""
-        params = self.upload_file(params)
-        artifact = Artifact(params)
+        metadata = self._upload_file(params)
+        artifact = self._create_artifact(params, metadata)
+        ImageRecognizer.auto_add_tags(artifact.elastic)
+        return respond_with(artifact.elastic), 200
 
+    def _create_artifact(self, params, metadata):
+        params.update(metadata)
+        artifact = Artifact(**params)
         artifact.save()
-        ImageRecognizer.auto_add_tags(artifact)
-        return respond_with(artifact), 200
+        return artifact
 
     @use_args(artifacts_validator.update_args())
     def update(self, params, object_id):
@@ -53,7 +57,7 @@ class ArtifactsController(ApplicationController):
         object_id = params.pop("id")
         try:
             artifact = Artifact.find(object_id)
-            artifact.update(params)
+            artifact.update(**params)
             return no_content()
         except NotFound:
             return {"error": "not found"}, 404
@@ -66,7 +70,7 @@ class ArtifactsController(ApplicationController):
             object_id = update_data.pop("id")
             try:
                 artifact = Artifact.find(object_id)
-                artifact.update(update_data)
+                artifact.update(**update_data)
             except NotFound:
                 return {"error": f"failed at <{object_id}>: not found"}, 404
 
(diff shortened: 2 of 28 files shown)


Pipfile
```diff
@@ -27,6 +27,7 @@ doublex-expects = "*"
 pep8 = "*"
 autopep8 = "*"
 rope = "*"
+PyHamcrest = "*"
```
Askir, line 30: These are the matchers used to match different Mock calls. So we don't have to use <ANY_ARGS> everywhere.

application/controllers/artifacts_controller.py
```diff
@@ -33,27 +34,32 @@ def show(self, object_id):
...
 
     @use_args(artifacts_validator.create_args())
     def create(self, params):
         """Logic for creating an artifact"""
-        params = self.upload_file(params)
-        artifact = Artifact(params)
+        metadata = type(self)._upload_file(params)
+        artifact = type(self)._create_artifact(params, metadata)
+        ImageRecognizer.auto_add_tags(artifact.elastic)
+        return respond_with(artifact.elastic), 200
```
Askir, line 46: Split into two methods and calling the wrapper instead of ElasticArtifact, everywhere that modifies values to keep both data models up to date. For now ElasticSearch will be the one used primarily though.

application/controllers/artifacts_controller.py
```diff
@@ -83,14 +89,39 @@ def delete(self, params, object_id):
...
+        file_saver.save()
+        return file_saver.get_metadata()
+
+
+class FileSaver:
+
+    def __init__(self, file):
+        self.file = file
+        self.file_date = None
+        self.filename = None
```
Askir, line 102: Saving the uploaded files is a little cleaner now.

application/controllers/teams_controller.py
```diff
@@ -28,7 +28,7 @@ class TeamsController(ApplicationController):
 
     def show(self, object_id):
         try:
-            team = Team.find_by(force=True, id_=object_id)
+            team = Team.find_by(force=True, id=object_id)
```
Askir: The new neo model uses id instead of id_. As Adrian said, overriding the id member is not a problem.

application/models/artifact.py
```diff
@@ -1,66 +1,45 @@
...
-        self.team_id = params.get("team_id", None)
-        if "score" in params:
-            self.score = params["score"]
+from .elastic_artifact import ElasticArtifact
+from .neo_artifact import NeoArtifact
+from .neo_tag import NeoTag
+import uuid
+
+
+class Artifact:
```
Askir, line 7: This is the wrapper for both the ElasticArtifact and the NeoArtifact. It isn't very pretty as I hope to get rid of it soon and do the syncing differently.

application/models/elastic_artifact.py
```diff
@@ -0,0 +1,66 @@
+""" Elastic search artifact model wraps api into crud methods"""
+from ..schemas.artifact_schema import ArtifactSchema
+from .esmodel import ESModel
+
+
+class ElasticArtifact(ESModel):
+    """ Handles saving and searching """
+
```
Askir, line 8: This is just the old Artifact. Now called ElasticArtifact to differentiate easier from NeoArtifact.

application/models/neo_artifact.py
```diff
@@ -0,0 +1,20 @@
+from .neo_ogm_model import NeoOGMModel
+from .neo_tag import NeoTag
+from py2neo.ogm import Property, RelatedTo
+
+
+class NeoArtifact(NeoOGMModel):
```
Askir: NeoArtifact class, with the new "custom" OGM.

application/models/neo_ogm_model.py
```diff
@@ -0,0 +1,88 @@
+import uuid
+from application.errors import NotFound
+from flask import current_app
+from py2neo.ogm import GraphObject, Property, Related, RelatedTo, RelatedFrom
+
+
+class NeoOGMModel(GraphObject):
```
Askir: OGM that wrapps the py2neo ogm to provide a subjectively nicer interface.

application/models/neo_tag.py
```diff
@@ -0,0 +1,23 @@
+from .neo_ogm_model import NeoOGMModel
+from py2neo.ogm import RelatedFrom, Property
+
+
+class NeoTag(NeoOGMModel):
```
Askir: NeoTags are the objects that keep the tag information. It has an extra method find_or_create_by as NeoTags should be unique in the database. I'll add a uniqueness validation aswell. So not using this method will throw exceptions if a Tag already exists.

application/models/team.py
```diff
@@ -1,90 +1,15 @@
...
-from py2neo import Node
-from ..errors import NotFound
-from ..schemas.team_schema import TeamSchema
+from application.schemas.team_schema import TeamSchema
+from .neo_ogm_model import NeoOGMModel
+from py2neo.ogm import Property
 
 
-class Team:
+class Team(NeoOGMModel):
```
Askir: Rewritten to use the OGM as well

specs/models/artifact_wrapper_spec.py
```diff
@@ -0,0 +1,135 @@
...
+from application.models.neo_artifact import NeoArtifact
+from application.models.elastic_artifact import ElasticArtifact
+from application.errors import NotInitialized
+from specs.spec_helpers import Context
+from specs.models.custom_matcher import have_node
+from specs.factories.elasticsearch import es_get_response
+from py2neo import Node
+import uuid
+
+with description('Artifact Wrapper') as self:
```
Askir, line 18: Tests for the wrapper. Not very pretty either.

specs/models/neo_artifact_spec.py
```diff
@@ -0,0 +1,90 @@
...
+from expects import expect, equal
+from application.models.neo_artifact import NeoArtifact
+from application.models.neo_tag import NeoTag
+from specs.spec_helpers import Context
+from specs.models.custom_matcher import be_uuid, have_node, have_relationship
+from flask import current_app
+from py2neo import Node
+
+node_name = "NeoArtifact"
+with description('NeoArtifact') as self:
```
Askir, line 11: Tests for the NeoArtifact. I think I haven't tested everything that the OGM provides. Not sure if we should have those though as they are provided by the base class?

application/controllers/artifacts_controller.py
```diff
@@ -10,21 +10,22 @@
...
     """ Controller for Artifacts """
 
     method_decorators = [check_es_connection]
 
     def show(self, object_id):
         """Logic for getting a single artifact"""
-
         try:
-            artifact = Artifact.find(object_id)
+            artifact = ElasticArtifact.find(object_id)
```
Adrian-St: Why do you use ElasticArtifact here? I though the idea was to 
always call the wrapper. Even if it just calls ElasticArtifact, makes it easyer to replace.

application/controllers/artifacts_controller.py
```diff
@@ -33,27 +34,32 @@ def show(self, object_id):
     def index(self, params):
         """Logic for querying several artifacts"""
 
-        artifacts = Artifact.search(params)
+        artifacts = ElasticArtifact.search(params)
```
Adrian-St: Same here

PR #52, opened 2019-01-26
Save resized versions of uploaded images

erksch:


application/controllers/artifacts_controller.py
```diff
@@ -7,6 +7,8 @@
 import werkzeug
 from flask import current_app
 from webargs.flaskparser import use_args
+from PIL import Image
+from pathlib import Path
```
erksch: Unnecessary.

application/controllers/artifacts_controller.py
```diff
@@ -83,6 +85,25 @@ def delete(self, params, object_id):
...
+        [prefix, suffix] = filename.rsplit('.', 1)
+        widths = [320, 480, 640, 750, 1080]
+
+        for width in widths:
+            width_percent = (width / float(image.size[0]))
+            height = int((float(image.size[1]) * float(width_percent)))
+            resized_image = image.resize((width, height), Image.ANTIALIAS)
+
+            new_filename = f"{prefix}_{width}w.{suffix}"
+            resized_image.save(new_filename)
```
Askir: You know exactly what I am going to say about this...
- too long  
- doesn't have anything to do with the controller
- should be a seperate class

For inspiration check my PR where i refactored the image upload. You might even wanna merge although there is a lot of different stuff in one PR again. Sorry about that!

Askir (changes requested):

Askir:
Oh also some tests would be nice IMO

application/controllers/artifacts_controller.py
erksch: Dat PR advertisement 😂

application/controllers/artifacts_controller.py
Adrian-St: Did you just review your own code? :D

Askir:
Rebased this for you, but you'll still have to do something about that big ass function.

application/controllers/artifacts_controller.py
erksch: yup

application/controllers/artifacts_controller.py
erksch: Please check my updates.

erksch:
What about testing here? I don't see tests for the "normal" image upload?

Askir:
Hmm on further inspection it actually looks like tests for the actual upload are missing, but we do have included a mock in the `spec_helper.py` checkout it out and see if you can work with that!

erksch:
E2E tests are failing for indeterministic reasons again :( 
Should not be caused by this PR.

application/artifacts/artifacts_view.py
```diff
@@ -179,16 +180,53 @@ def patch(self, **params):
...
+    def _generate_file_name(self, width):
+        """Generate filename for a resized image of given width"""
+        [prefix, suffix] = self.original_file_name.rsplit('.', 1)
+        filename = f"{prefix}_{width}w.{suffix}"
+        return filename
+
+    def _calculate_height(self, width):
+        """Calculate height of resized image with given width"""
+        width_percent = (width / float(self.image.size[0]))
+        height = int((float(self.image.size[1]) * float(width_percent)))
```
Askir: Do you not want to round this instead of always flooring it maybe?
That one pixel probably doesn't matter though I guess :D

specs/requests/artifacts_spec.py
```diff
@@ -102,6 +112,32 @@
                 with it("returns an empty list for tags"):
                     expect(self.response.json.get("tags")).to(equal([]))
 
+                with it("saves the image to the configured uploads directory"):
+                    file_name = self.response.json.get("url").split("/")[-1]
+                    file_path = os.path.join(current_app.config["UPLOAD_FOLDER"], file_name)
```
Askir, line 117: You are replicating a lot of application logic in these tests.
The file_name should not be retrieved from the response imo. And the file_path should be hardcoded as well.

specs/requests/artifacts_spec.py
```diff
@@ -102,6 +112,32 @@
...
+                    expect(os.path.isfile(file_path)).to(equal(True))
+
+                with description("resized image versions"):
+                    with before.each:
+                        self.widths = [320, 480, 640, 750, 1080]
+
+                        original_file_name = self.response.json.get("url").split("/")[-1]
+                        [self.prefix, self.suffix] = original_file_name.rsplit('.', 1)
```
Askir, line 125: Again the original file_name should not be calculated from the response.

specs/requests/artifacts_spec.py
```diff
@@ -102,6 +112,32 @@
...
+                            file_path = os.path.join(current_app.config["UPLOAD_FOLDER"], file_name)
+                            expect(os.path.isfile(file_path)).to(equal(True))
+
+                    with it("saves resized versions with correct widths"):
+                        for width in self.widths:
+                            file_name = f"{self.prefix}_{width}w.{self.suffix}"
+                            file_path = os.path.join(current_app.config["UPLOAD_FOLDER"], file_name)
+                            image = Image.open(file_path)
+                            expect(image.width).to(equal(width))
+                            image.close()
```
Askir, line 139: Otherwise pretty good testing though!

Askir (approved):

Askir:
~~Actually I checked how long uploading takes now with the resizing @erksch :~~
timetoresize

~~Up to 5 seconds here. No wonder E2E fails with a timeout of 3 seconds. Two solutons one bandaid one propper:~~
1. ~~Set the timeout higher (6-7 seconds maybe)~~
2. ~~Put the resizing and/or image saving into a separate thread.~~

Askir:
Actually nevermind the timeout is set to 30.000 ms not 3.000. I wonder why it crashes then... Has been running butter smooth on my refactoring branch. And nothing has been merged in the front end since. But the error occurred once on master already so its not because of the resizing.

erksch:
@Askir I merged a fix for the e2e tests on the frontend a couple of days ago. 
So it works properly now, I guess.

application/artifacts/artifacts_view.py
erksch: Haha this took me a while :D 
I think you read that wrong. I am converting to float not flooring.

erksch: But thinking about it all the float casts are pretty much useless.
I refactored to:
```
width_percent = width / self.image.size[0]
height = int(self.image.size[1] * width_percent)
```

erksch: I'll rename `width_percent` to `width_ratio`.

specs/requests/artifacts_spec.py
erksch: The problem is that I don't know the file name if I don't retrieve it from the response :(
The backend saves it on a very different name so there is no chance to find the file without checking the response.

specs/requests/artifacts_spec.py
erksch: Same here. The original filename is the name Elija assigned to the file.

specs/requests/artifacts_spec.py
erksch: Why hard code the file_path when the upload folder is configurable. It would just break the tests if you use a rename it.

specs/requests/artifacts_spec.py
erksch: Thank you!

PR #53, opened 2019-01-27
Files volume for docker-compose 

erksch:


.env.dc
```diff
@@ -1,6 +1,6 @@
 ES_URL=http://elasticsearch_dev:9200
 NEO4J_URL=bolt://neo4j_dev:7687
 FILE_SERVER=http://localhost:8080
-UPLOAD_FOLDER=uploads/
+UPLOAD_FOLDER=uploads
```
erksch, line 4: It is important to remove the dangling slash here because docker-compose would not mount the volume properly with it. 
The change does not break the normal upload workflows in the code!

Adrian-St:
Does this only work when starting elija in docker?
I tried it out with elija running locally and it didn't work

erksch:
It's important to remove the slash from your UPLOAD_FOLDER variable in your env file!
See .env.example.

PR #55, opened 2019-01-28
Add shovel.

Askir:
This allows you to run migrations and other setup tasks by doing:
`shovel utils.setup_neo_constraints` for example. We should write some scripts that execute everything necessary to setup both databases.


erksch (approved):
I'm hyped to see more scripts for migrations etc.

PR #57, opened 2019-01-31
Add and configure poetry for project.

Askir:
As to why checkout these resources:
- https://www.reddit.com/r/Python/comments/8jd6aq/why_is_pipenv_the_recommended_packaging_tool_by/
- https://chriswarrick.com/blog/2018/07/17/pipenv-promises-a-lot-delivers-very-little/

I also had problems with pipenv because `pipenv run` automatically loads the .env file. Which prevents `pipenv run mamba specs` to load the env.testing. You can disable that with an ENV variable but that doesn't seem to be a clean solution for me.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,22 +11,16 @@ jobs:
           NEO4J_dbms_memory_pagecache_size: 100M
     steps:
       - checkout
-      - restore_cache:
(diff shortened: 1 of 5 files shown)


erksch (approved):
I would delete the pipenv stuff later.
Maybe it's good for migration purposes to keep it now 
but having two package and env managers is baad.

PR #58, opened 2019-01-31
Add annotations to differentiate from tags.

arne-z:


application/recognition/image_recognition.py
```diff
@@ -51,22 +54,22 @@ def _call_google_api(cls, artifact):
...
+                    text_annotations.append(text['description'])
+        return label_annotations, text_annotations
 
     @classmethod
-    def add_tags_artifact(cls, artifact, new_tags):
-        existing_tags = artifact.tags or []
-        tags = existing_tags + new_tags
-        artifact.update({"tags": tags})
+    def add_tags_artifact(cls, artifact, label_annotations, text_annotations):
+        artifact.update({'label_annotations': label_annotations,
```
Askir, line 75: Are you certain this works? Because previously you were given an ESArtifact here so these methods are proably not available for you at this point.

application/models/neo_artifact.py
```diff
@@ -9,6 +9,8 @@ class NeoArtifact(NeoModel):
     __primarylabel__ = "NeoArtifact"
     url = Property()
     tags = RelatedTo(NeoTag, "TAGGED_WITH")
+    label_annotations = RelatedTo(NeoTag, "ANNOTATED_WITH_LABEL")
+    text_annotations = RelatedTo(NeoTag, "ANNOTATED_WITH_TEXT")
```
Askir, line 14: So we have three different kinds of tags? User tags, google created labels and google created text tags?
Seems fine to me, I'm not sure if this is the correct or best way to model it but lets just do it and if it's stupid we at least learned something.

Other options would still be:
- Double relations (every tag with TAGGED_WITH + special relation like ANNOTATED_WITH_TEXT).
- Single TAGGED_WITH relation + properties on the relation (but the support for that in py2neo seems a little lackluster).

application/recognition/image_recognition.py
arne-z: ` artifact.update({"tags": tags})` certainly works. Updating with LabelAnotations and TextAnotations should work here since I changed the update method of artifact (see artifact.py line 46).

arne-z: "should work" - this obviously means I have not tested enough. I will come back when that is tested^^

Askir:
I'm afraid we'll have to redo this, checkout https://github.com/hpi-sam/ask-your-repository-api/pull/61
And https://blog.askir.me/2019/02/04/new-libraries-and-tools-for-our-backend/ on why

PR #61, opened 2019-02-05
Use Neomodel instead of py2neo. [FEEDBACK WANTED]

Askir:
This is how Neomodel would work. 

But in general I think the library is quite a bit better. It also has the ability to have Models for Relationships so we are able to save properties in the "TAGGED_WITH" Relationship for example. Which is really nice imo and was definitely missing in py2neo. 

The interface isn't perfect. For some reason the interal neo id is exposed as .id which is really annoying because now we have to use id_ again to save our own uuid. But at least they provide an easy property for that yay!

Some other things that made me stumble:
- The uuid is saved as hex (without the hypens "-") but when exported via Marshmallow the hypens are added again... So the returned id cannot be searched for in the database. you have to remove the hyphens first.
- All `find` or `all` methods we used before are via the interface `object.nodes.xyz` which again doesn't hide the graph very well. But I think we can provide our own wrappers for that again if we want to.
- In general the interace is more adapted to django than rails so I am not quite as familar with it but I guess @arne-z is going to like it.


Askir:
Pipeline broke I know it's just a spike PR so everyone can see how it would look like. I didn't delete all references to py2neo yet sorry.

Askir:
@erksch would there be any problems in the frontend if from now on we would ommit the hyphens in uuids?

Askir:
E2E is fixed with this PR in Jona: https://github.com/hpi-sam/ask-your-repository-web/pull/42 just waiting for eriks approval.

I know this is a huge PR again but I'd really appreciate reviews at least on a basic level. To make sure everybody understands what this was about and also to make sure I didn't miss any obvious problems

erksch:
@Askir There would be now problem whatsoever in the frontend. If the ID is consistent in all resources everything will work

.circleci/config.yml
```diff
@@ -4,7 +4,7 @@ jobs:
     docker:
       - image: circleci/python:3.7
         environment:
-          NEO4J_URL: bolt://localhost:7687
+          NEO4J_URL: bolt://:@localhost:7687
```
erksch, line 7: What is that?

.env.testing
```diff
@@ -1,5 +1,5 @@
 ES_URL=http://localhost:19200
-NEO4J_URL=bolt://localhost:17687
+NEO4J_URL=bolt://xd:dd@localhost:17687
```
erksch: It gets even worse :D whats that?

application/__init__.py
```diff
@@ -31,8 +31,9 @@ def create_app(config_filename=None):
     app.config.from_pyfile(config_filename)
     app.es = (Elasticsearch(app.config['ELASTICSEARCH_URL'])
               if app.config['ELASTICSEARCH_URL'] else None)
-    app.graph = (Database(app.config['NEO4J_URL']).default_graph
-                 if app.config['NEO4J_URL'] else None)
+    config.ENCRYPTED_CONNECTION = False
```
erksch: Why not encrypted?

application/controllers/artifacts_controller.py
```diff
@@ -11,27 +11,29 @@
...
 from ..synonyms.synonyms import SynonymGenerator
 from ..validators import artifacts_validator
 
+
 @socketio.on("SYNCHRONIZED_SEARCH")
 @socketio_args(artifacts_validator.search_args())
 def synchronized_search(params):
     """ Called from client when presentation mode is on """
-    artifacts = Artifact.search(params)
+    artifacts = ArtifactBuilder.search(params)
```
erksch, line 29: Why is it called ArtifactBuilder now?

application/controllers/artifacts_controller.py
```diff
@@ -69,11 +74,12 @@ def create(self, params):
...
         artifact = self._create_artifact(params, metadata)
         ImageRecognizer.auto_add_tags(artifact.elastic)
-        return respond_with(artifact.elastic), 200
+        return respond_with(artifact.neo), 200
 
     def _create_artifact(self, params, metadata):
         params.update(metadata)
-        artifact = Artifact(**params)
+        artifact = ArtifactBuilder()
+        artifact.build_with(**params)
```
erksch, line 82: I get that you want the ArtifactBuilder to be a builder pattern for the elastic and neo models, but having this and the rails like Model methods (like find_by etc.) on that Builder as well seems a bit unintuitive. 
So maybe don't call it builder or extract the builder part from the model wrapper part.

application/models/artifact.py
```diff
@@ -1,57 +1,37 @@
...
+""" Access to Artifacts via Neo4J """
+
+from neomodel import (StructuredNode, StringProperty, DateTimeProperty,
+                      RelationshipTo, RelationshipFrom)
+
+from application.models.mixins import DefaultPropertyMixin, DefaultHelperMixin
+from application.schemas.artifact_schema import NeoArtifactSchema
+
+
+class Artifact(StructuredNode, DefaultPropertyMixin, DefaultHelperMixin):  # pylint:disable=abstract-method
```
erksch, line 10: Shouldn't this be called NeoArtifact?

erksch (changes requested):

erksch:
@Askir Great work overall! Well tested imo.

.circleci/config.yml
Askir: Instead of URLs Neomodel uses these Database access links, which are similar to rails database links.
You can check them out here: https://apidock.com/rails/ActiveRecord/Base/establish_connection/class
Here is an example for rails from that site: "postgres://myuser:mypass@localhost/somedatabase".
Explanation is protocol://username:password@url:port
But since we have authentication disabled in CI and development I just left username and password empty.

.env.testing
Askir: Good catch. I didn't know if leaving empty worked so I added random shit here. gonna remove that :D

application/__init__.py
Askir: Is a pain to setup for development and our database is not available remotely anyways.

application/controllers/artifacts_controller.py
Askir: The endstate I want that class to be is an ArtifactBuilder. It should encapsulate how the relations between artifacts, tags and teams are connected. Currently it also does the job of syncing with elastic search which is something I'll be working on next. So the class only has one responsibility left.

application/controllers/artifacts_controller.py
Askir: Well, as I said the class is in a very rough state since it has at least two responsibilities right now. I think it will get much more clear once I extract all the ES stuff from it. Then all the find_by calls for example can also be directly executed on the Artifact and not the ArtifactBuilder. You are right though that the name is not very expressive right now. It's difficult to find one that explains all the responsibilites right now though...

application/models/artifact.py
Askir: I decided to just call all the models without Neo now. Since those are our default models. I also think Elasticsearch doesn't really need models in the future so having these called Neomodels is just going to lead to confusion with the library name :D

application/__init__.py
Askir: Actually nevermind, the problem seems to have been something else. I enabled it again :)

application/controllers/artifacts_controller.py
Askir: My overall idea is to:
- Have Artifacts be accessed directly for index/show routes
- Have the artifact builder be a small helper class for creating Artifacts (and maybe updating)
- Have a new class ElasticSearchSynchronizer, that manages populating elasticsearch data. And is called in a post_save hook on the Neo4J models.

I want to do all three in a seperate PR since this one is already a pain to read and very large :D

.circleci/config.yml
erksch: Ah ok I get it.

erksch (approved):

PR #63, opened 2019-02-15
Search fixes

Askir:
default indices were somehow wrong (thought we fixed that ages ago.)
And I added the url with http://localhost into elasticsearch and readded it after retrieving agains so urls looked like this:
'http://localhosthttp://localhost/real_image.png'

### DIFF
--- a/shovel/utils/setup_indices.py
+++ b/shovel/utils/setup_indices.py
@@ -15,8 +15,8 @@ def setup_indices():
             "image": {
                 "properties": {
                     "tags": {"type": "text"},
-                    "file_url": {"type": "text"},
-                    "team_id": {"type": "text"},
+                    "file_url": {"type": "keyword"},
+                    "team_id": {"type": "keyword"},
                     "file_date": {
                         "type": "date",
                         "format": "strict_date_optional_time||epoch_millis"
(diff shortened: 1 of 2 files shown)


shovel/utils/setup_indices.py
```diff
@@ -15,8 +15,8 @@ def setup_indices():
             "image": {
                 "properties": {
                     "tags": {"type": "text"},
-                    "file_url": {"type": "text"},
-                    "team_id": {"type": "text"},
+                    "file_url": {"type": "keyword"},
+                    "team_id": {"type": "keyword"},
```
erksch, line 19: What's up with that?

Askir: basically the mappings are elasticsearches "Schema" so team_id had the type text instead of keyword.
Text is good because it supports full text search with partial matchin etc. but it doesn't support filtering, sorting and other more "aggregational functions". So for IDs, emails, etc. you should use the type keyword instead of text. We aren't currently using the file_url for anything but if we were I think it should be a keyword as well.

Also btw this was already done inside the migrations folder so if you ran setup indices + migrate everything would've already worked. But why keep the old schema here? doesn't make any sense to me.

erksch: Ok thanks for the explanation!

PR #64, opened 2019-02-20
Notify dialogflow adapter of team creation.

henleo:


.env.testing
```diff
@@ -1,5 +1,6 @@
 ES_URL=http://localhost:19200
 NEO4J_URL=bolt://:@localhost:17687
+DIALOGFLOW_ADAPTER=http://www.localhostdrolf.de
```
Askir: Some url that tells you what is wrong would be nice here imo. i.e. `tobito.should.be.mocked`

application/controllers/teams_controller.py
```diff
@@ -1,10 +1,13 @@
 """
 Handles all logic of the artefacts api
 """
+import os
+import requests
+
+from webargs.flaskparser import use_args
+#from eventlet import spawn_n
```
Askir: remove ? :D

application/controllers/teams_controller.py
```diff
@@ -44,6 +47,9 @@ def index(self, params):  # pylint: disable=W0613
     def create(self, params):
         """Logic for creating a team"""
         team = Team(name=params["name"]).save()
+        print(team)
```
Askir: remove? :D

application/controllers/teams_controller.py
```diff
@@ -44,6 +47,9 @@ def index(self, params):  # pylint: disable=W0613
...
+        self.notify_of_team_creation(team)
+        #spawn_n(self.notify_of_team_creation, team)
```
Askir: remove? :D or maybe just uncomment if you want to make this threaded now, see other comments

specs/requests/teams_spec.py
```diff
@@ -49,6 +51,24 @@
             with it('saves the team'):
                 expect(Team.nodes.get_or_none(id_=self.response.json["id"], name="My Team")).to_not(be(None))
 
+        with description('notify dialogflow adapter'):
+            with before.each:
+                current_app.config['DIALOGFLOW_NOTIFY'] = True
+                self.DIALOGFLOW_URL = os.environ.get('DIALOGFLOW_ADAPTER') + '/teams'
+                self.rsps = responses.RequestsMock()
+                self.rsps.__enter__()
```
Askir: This is honestly pretty hacky. If you want to leave it this way at least leave a comment on how it works. Or change it back to: `with responses.RequestsMock() rsps:`

Askir (commented):
Aside from a few comments, looks pretty functional. Wont block merge, I think you'll know what to fix. Definitely should implement an observer like pattern for these sort of structures in the future though! If you're a real busy bee you could already create a refactoring issue for that :)

Askir:
Oh also one more thing! please squash the "Fix linting" and similar commits into one.

Adrian-St:
Changes have been merged in another branch

PR #65, opened 2019-02-20
User management

Adrian-St:


application/controllers/users_controller.py
```diff
@@ -0,0 +1,91 @@
...
+from .application_controller import ApplicationController
+from ..models.user import User
+from ..responders import respond_with, no_content
+from ..validators import users_validator
+from ..extensions import bcrypt
+
+class UsersController(ApplicationController):
+    """ Controller for users """
+
+    @jwt_required
```
Askir, line 16: okay so I checked the docs. This checks if a jwt token is present. But does it check if it belongs to the user that is trying to be accessed? How do we configure pages and resources to only be avaiable for one person. So that for example a user can only see his own preferences page but not that of others?

application/controllers/users_controller.py
```diff
@@ -0,0 +1,91 @@
...
+                or User.find_by(email=params["email_or_username"], force=False))
+        if not user or not bcrypt.check_password_hash(user.password, params["password"]):
+            return {"error": "Bad username or password"}, 401
+        access_token = create_access_token(identity=user.id_)
+        response = respond_with(user)
+        response["token"] = get_csrf_token(access_token)
+        response = jsonify(response)
+        set_access_cookies(response, access_token, 1000000)
+        response = make_response(response, 200)
+        response.mimetype = 'application/json'
```
Askir: This could use some simple method extraction to make it more readable:
```
user = self.find_user()
if not self.validate_user(user):
  return ...
return self.build_login_response_with(create_access_token(...))
```

application/models/team.py
```diff
@@ -11,3 +11,4 @@ class Team(StructuredNode, DefaultPropertyMixin, DefaultHelperMixin):  # pylint:
     name = StringProperty(required=True)
 
     artifacts = RelationshipTo('.Artifact', 'UPLOADED')
+    users = Relationship('.User', 'MEMBER')
```
Askir: would prefer this to be a directed relationship so Team -[HAS_MEMBER]-> User or User-[IS_MEMBER]-> Team.
That makes the graph more readable and I think it is less error prone when traversing the graph later.

application/routes.py
```diff
@@ -13,11 +13,15 @@ def create_routes(api):
...
     from .controllers.teams_controller import TeamsController
+    from .controllers.users_controller import UsersController
 
     add_method(api, '/', "show_status", HomeController, method="get")
     add_resource(api, '/images', ArtifactsController)
     add_method(api, '/images', "update_many", ArtifactsController, method="patch")
     add_resource(api, '/teams', TeamsController)
+    add_resource(api, '/users', UsersController)
+    add_method(api, '/users/login', "login", UsersController, method="post")
+    add_method(api, '/users/logout', "logout", UsersController, method="post")
```
Askir: Do these have to work like this. Didn't we want to make these routes restful with an `/authentication` endpoint that you can post and delete to?

application/schemas/user_schema.py
```diff
@@ -0,0 +1,17 @@
+""" Defines schema for database Team objects """
+from marshmallow import fields
+
+from ..base import BaseSchema, output_decorator
+
+
+class UserSchema(BaseSchema):
+    """ Schema for importing and exporting Team objects """
+    id_ = fields.UUID(missing=None)
```
Askir, line 9: id_ is actually saved simply as a string. I'm not sure if that might mess with the schema later. You might wanna consider just making this a string with a `validate_uuid` hook (same in all other model schemas then though)

application/schemas/user_schema.py
```diff
@@ -0,0 +1,17 @@
...
+    username = fields.String()
+    email = fields.String()
+
+    @output_decorator
+    def transform_fields(self, data):
+        """ Transforms field for output """
+        data["id"] = data.pop("id_")
+        return data
```
Askir, line 17: we should provide a mixin for this

instance/production_config.cfg
```diff
@@ -0,0 +1,31 @@
+import os
```
Askir, line 1: What actually decides which config is being used? Don't we have to change our build process so that production_config.cfg will be used?

specs/spec_helpers.py
```diff
@@ -1,11 +1,32 @@
...
+        server_name = current_app.config["SERVER_NAME"] or "localhost"
+        self.set_cookie(server_name,
+                        config.access_cookie_name,
+                        value=access_token,
+                        max_age=config.cookie_max_age,
+                        secure=config.cookie_secure,
+                        httponly=True,
+                        domain=config.cookie_domain,
+                        path=config.access_cookie_path)
+        return get_csrf_token(access_token)
```
Askir: I don't quite understand what you are doing here. Is this some sort of mock for the login? If yes, why does it have to be mocked, if not what exactly is this? :D

Askir:
Looks promising just need some clarification on a few topics.

application/controllers/users_controller.py
Adrian-St: the user_id would be decoded in the jwt token for permission control

specs/spec_helpers.py
Adrian-St: I think setting of cookies didn't work properly with the normal login route, that's why I wrote that method
that calls set_cookie of the testing client directly. However, it seems to work with the normal route now, so I will use that instead :D

PR #70, opened 2019-03-01
Annotations with Neomodel.

arne-z:


application/models/artifact.py
```diff
@@ -37,6 +40,30 @@ def tags_list(self):
...
+            tags_list.append(tag.name)
+        return tags_list
+
+    @property
+    def text_tags_list(self):
+        """ Returns this artifacts text recognition generated tags as string array """
+        tags_list = []
+        for tag in self.text_tags:  # pylint:disable=not-an-iterable
+            tags_list.append(tag.name)
+        return tags_list
```
Askir: this is all but pretty now, we either need to be able to create these getters dynamically. Or find a different way to make the iterators accessible for the schema.

One thing that definitly makes this all a bit better already though would simply be `return list(self.tags)` but even that is too much duplication imo.

application/models/artifact_builder.py
```diff
@@ -19,26 +20,16 @@ def __init__(self):
 
     def build_with(self, **properties):
         """ builds self.neo and saves it"""
-        neo_properties = self._make_properties_neo_compatible(**properties)
+        props = PropertyBuilder(**properties)
+        neo_properties = props.node_properties
+        print(neo_properties)
```
Askir: remove plz :D

application/models/artifact_builder.py
```diff
@@ -19,26 +20,16 @@ def __init__(self):
...
         self.neo = Artifact(**neo_properties).save()
-        if 'tags' in properties:
-            self._add_tags_to_neo(properties['tags'])
-        if 'team_id' in properties and properties.get('team_id') is not None:
-            self._connect_to_team(properties['team_id'])
-        properties['id'] = self.neo.id_
+        print(self.neo)
```
Askir: also remove plz :D

application/models/artifact_builder.py
```diff
@@ -19,26 +20,16 @@ def __init__(self):
...
+        self._set_full_tags(props.tag_properties, override_tags=False)
+        if 'team_id' in props.properties and props.properties.get('team_id') is not None:
+            self._connect_to_team(props.properties['team_id'])
```
Askir: A better structuring of all this would be a classification into `node_properties` and `relationship_properties` And then have a ArtifactNodeBuilder (or just a method for that since it is mostly constructor calling) and have an ArtifactRelationshipBuilder  that works with the relationship_properties and adds the tags, team and in the future author via different methods.

Thoughts on that design?

application/models/artifact_builder.py
```diff
@@ -49,20 +40,50 @@ def _add_tags_to_neo(self, tags):
             self.neo.tags.connect(created_tag)
         self.neo.save()
 
-    def _overwrite_tags_in_neo(self, tags):
-        self.neo.tags.disconnect_all()
-        self._add_tags_to_neo(tags)
+    def _add_user_tags_to_neo(self, user_tags):
+        for tag in user_tags:
+            print(self.neo)
+            print(self.neo.user_tags)
```
Askir: more prints to remove! :D

application/models/artifact_builder.py
```diff
@@ -49,20 +40,50 @@ def _add_tags_to_neo(self, tags):
...
+        for tag in label_tags:
+            created_tag = Tag.find_or_create_by(name=tag)
+            self.neo.label_tags.connect(created_tag)
+        self.neo.save()
+
+    def _add_text_tags_to_neo(self, text_tags):
+        for tag in text_tags:
+            created_tag = Tag.find_or_create_by(name=tag)
+            self.neo.text_tags.connect(created_tag)
+        self.neo.save()
```
Askir, line 45: These methods can easily be dried by giving them an extra parameter `tagType`

Askir (changes requested):
good start for sure, could use a bit more refactoring and a few more tests as well It think. The ArtifactBuilder seems to start to become a core component of the backend so it should be tested well and we have to think about the code design in more detail.

application/models/artifact_builder.py
arne-z: done

application/models/artifact_builder.py
arne-z: later, once testing etc is done. currently needed so I can see what is happening^^

application/models/artifact_builder.py
arne-z: Sounds good, look at the new property_handler whether you like the changes.

application/models/property_builder.py
```diff
@@ -0,0 +1,46 @@
...
+class PropertyBuilder:
+    """Describe a PropertyBuilder."""
+
+    def __init__(self, **properties):
+        self.__properties = properties
+        self.__relationship_properties = None
+        self.__node_properties = None
+        self.relationship_property_types = ['tags', 'user_tags', 'label_tags', 'text_tags',
+                                            'team_id']
+        self.node_property_types = ['file_url', 'file_date']
```
Askir: Can we have these two instance variables at the top as constants? Just as capslock class variable for example.

application/models/tag.py
```diff
@@ -1,11 +1,21 @@
...
 from neomodel import StructuredNode, StringProperty, RelationshipFrom
+
 from application.models.mixins import DefaultPropertyMixin, DefaultHelperMixin
 
 
-class Tag(StructuredNode, DefaultPropertyMixin, DefaultHelperMixin):  # pylint:disable=abstract-method
+# pylint:disable=abstract-method
+class Tag(StructuredNode, DefaultPropertyMixin, DefaultHelperMixin):
     """ The class that manages Tags """
     # schema = ArtifactSchema does not exist yet
```
Askir: oh there is a lost comment here. Feel free to remove :D

specs/models/relations/artifact_wrapper_spec.py
```diff
@@ -62,16 +63,16 @@
                 self.artifact = ArtifactBuilder()
                 self.artifact.build_with(file_url='asdf', type='image')
                 self.artifact.save()
-                self.artifact.update_with(tags=['a', 's', 'd', 'f'])
+                self.artifact.update_with(user_tags=['a', 's', 'd', 'f'])
```
Askir: Do you mind adding some tests for the other tag types as well here?

Askir (approved):
Some specific tests for the new relations would be nice. Otherwise looks pretty good.
Definitely have to do the schema refactoring though to make the property builder cleaner and elastic syncing less error prone.

application/models/property_builder.py
arne-z: done

application/models/tag.py
arne-z: # schema = ArtifactSchema does not exist yet
 this one you meant right? will remove it.

specs/models/relations/artifact_wrapper_spec.py
arne-z: The other tag types are actually tested in the image_recognition_spec.py which is where they are used as well.

arne-z: done though.

PR #72, opened 2019-03-02
Give teams members.

Askir:


application/controllers/teams_controller.py
```diff
@@ -4,21 +4,22 @@
...
 from ..extensions import socketio
-from ..models.team import Team
+from ..models import Team, User
 from ..responders import respond_with
 from ..validators import teams_validator
 
-@socketio.on("JOIN_TEAM")
-def on_team_join(data):
+@socketio.on("ENTER_TEAM_SPACE")
+def on_enter_team_space(data):
```
Askir, line 16: This will break presentation mode again as long as it isn't fixed on the front end as well I think?
I thought Luise didn't push this but it's probably faster to change it in the front end as well than removing it here again.

application/controllers/teams_controller.py
```diff
@@ -4,21 +4,22 @@
...
     """ Logic for connecting to a Team with socketio """
     current_app.logger.info("Join Room: " + data["team_id"])
     join_room(str(data["team_id"]))
 
-@socketio.on("LEAVE_TEAM")
-def on_team_leave(data):
+@socketio.on("EXIT_TEAM_SPACE")
+def on_exit_team_space(data):
```
Askir, line 22: same here.

application/base.py
```diff
@@ -24,7 +24,7 @@ def decorate(instance, data):
 class BaseSchema(Schema):
     """ Initialize Schemas with Schema(Model) """
 
-    def __init__(self, model, *args, decorate=False, create_objects=True, **kwargs):
+    def __init__(self, *args, model=None, decorate=False, create_objects=True, **kwargs):
```
Askir, line 27: This being a required argument messed with Nested fields. So I made a keyword argument out of it. @Adrian-St note for you and me for the future. When building wrappers don't make arguments required and always provide default values.

application/controllers/teams_controller.py
erksch: For migrating purposes it would be better to have both and remove the JOIN_TEAM when the frontend supports ENTER_TEAM_SPACE.

erksch: But I guess this whole PR is going to mess with the frontend and both have to be updated rather simultaneously.

erksch (approved):

erksch:
Very very nice!

PR #74, opened 2019-03-03
Delete file on /images DELETE

Askir:


erksch:
Nice that you already considered different image sizes!

PR #76, opened 2019-03-04
Add user to artifact

Adrian-St:
Connects the user to the artifact if the user is logged in.
Otherwise they don't connect and the user_id is none.


henleo (changes requested):
sudo test for persistence

henleo:
sudo write contract test

luisebenkert (approved):

PR #83, opened 2019-03-13
Remove Flask Restful and refactor routing

Askir:
We're removing flask restful in favor of flask-apispec which is a more maintained and up to date library. That also uses other libraries that we've already implemented.

I'm trying to keep close to the structure of this example project: https://github.com/gothinkster/flask-realworld-example-app/tree/master/conduit

Thinking about changing the folder structure accordingly as well. Thoughts about that?
So having and artifacts, tags, users, etc. folder with the View/Controller, Model and Schemas/validators.

Small hint: It's probably useful to review this PR in split view instead of unified:
splitinsteadofunified

I also decided that for our usecase a divisional structure makes more sense than a functional one. I talked to @Adrian-St about this. I hope everyone else agrees: http://exploreflask.com/en/latest/blueprints.html#divisional


application/artifacts/artifact_schema.py
```diff
@@ -34,3 +34,14 @@ def build_url(file_url):
...
         return current_app.config["FILE_SERVER"] + \
                "/" + file_url
+
+    @post_dump(pass_many=True)
+    def dump_artifacts(self, data, many):
+        """ add a key for the returned collection """
+        if many:
+            return {'images': data, 'images_count': len(data)}
+        return data
+
```
Adrian-St, line 44: Sweet, a shame that this doesn't work with flask apispec

application/artifacts/artifacts_view.py
```diff
@@ -7,22 +7,23 @@
...
+from application.artifacts.artifact_schema import ARTIFACT_SCHEMA, ARTIFACTS_SCHEMA
+from application.artifacts.elastic import ElasticSearcher
+from application.artifacts.image_recognition import ImageRecognizer
+from application.artifacts.synonyms import SynonymGenerator
+from application.errors import check_es_connection
+from application.extensions import socketio
+from application.responders import respond_with, no_content
+from application.socketio_parser import use_args as socketio_args
+from application.teams.team import Team
+from .artifact import Artifact
```
Adrian-St, line 27: Any reason you don't use relative imports anymore?
I think its nice to see quickly which modules belong to this Blueprint and which are external.

application/artifacts/artifacts_view.py
```diff
@@ -69,21 +70,70 @@ def _find_multiple_by(params):
     return artifacts.order_by('created_at')[_from:_to]
 
 
-class ArtifactsController(ApplicationController):
-    """ Controller for Artifacts """
-
-    method_decorators = [check_es_connection]
+class ArtifactView(MethodResource):
+    """ Access Artifacts by id """
```
Adrian-St: Just a reminder: Docstrings start directly after the """ so you would have a space here which isn't good practice.
Obviously I don't expect you to change all of them, but if you change old docstrings it would be useful ;)

.pylintrc
```diff
@@ -315,7 +315,7 @@ indent-after-paren=4
 indent-string='    '
 
 # Maximum number of characters on a single line.
-max-line-length=100
+max-line-length=120
```
Adrian-St, line 318: Is that really necessary? Pep suggests maximum line length of 80

application/artifacts/artifacts_view.py
```diff
@@ -69,21 +70,70 @@ def _find_multiple_by(params):
...
+    def patch(self, **params):
+        """Logic for updating an artifact"""
+        try:
+            artifact = Artifact.find_by(id_=params.pop("id"))
+            builder = ArtifactBuilder.for_artifact(artifact)
+            builder.update_with(**params)
+            return no_content()
+        except Artifact.DoesNotExist:
+            return abort(404, 'artifact not found')
+
```
Adrian-St, line 103: I'm not the biggest fan of copy pasting everything for put and patch when it's the same.
I would either move the method into update and call that from both (you could even change the name to update and 
define two lambdas for PUT and PATCH that pass the call to update) or override the MethodView class to call PUT when PATCH isn't defined and vice versa.

application/routes.py
```diff
@@ -1,30 +1,22 @@
...
-    add_resource(api, '/teams', TeamsController)
-    add_resource(api, '/users', UsersController)
-    add_resource(api, '/authentications', AuthenticationsController, only=["create"])
-    add_method(api, '/authentications', "delete", AuthenticationsController, method="delete")
-    add_resource(api, '/presentations', PresentationsController, only=["create"])
-    add_method(api, '/images/<object_id>/tags', "add_tags",
-               TagsController, method="post")
-    add_method(api, "/tags/suggested", "suggested_tags",
-               TagsController, method="get")
+    app.add_url_rule('/tags/suggested', "suggested_tags", TagsView().suggested_tags)
```
Adrian-St, line 21: I think we should move that to teams btw, but not in that PR.

application/base.py
```diff
@@ -49,121 +47,3 @@ def handle_errors(schema, errors, obj):  # noqa
...
-    @classmethod
-    def new_resource(cls, name, method):
-        """ Creates resource class for new method """
-
-        methods = {}
-        methods["method_decorators"] = cls.method_decorators
-        methods[method] = getattr(cls(), name)
-        new_class = type(cls.__name__ + name.capitalize(), (Resource,), methods)
-
-        return new_class
```
Adrian-St, line 169: Goodby Python on segways :cry:

application/base.py
```diff
@@ -49,121 +47,3 @@ def handle_errors(schema, errors, obj):  # noqa
                      .format(errors))
     current_app.logger.error(error_message)
     raise ValueError(error_message)
```
Adrian-St: Is that still used? If yes we should move it into a different file, base.py is a horrible name :D

Adrian-St: Ah okay that's where the base schema still lives. I think base_schema.py would be better but we can ignore that for now

Adrian-St (changes requested):
looks great :) Only minor remarks / questions

application/artifacts/artifacts_view.py
Askir: I'm using relative imports whenever I don't have to use `..` to go up a folder. Feels not very intuitive to me and doesn't give a good overview.
This way you can see everything starting with a simple dot is connected and an internal requirement of this module. While everything starting with application is external.

application/artifacts/artifacts_view.py
Askir: Yeh I'm getting confused by old code and old docstrings I can change it project wide with a find and replace though.

.pylintrc
Askir: Our prospector.yaml has pep8 set to check for 120 and the .pylintrc had 100 (which lead to no pep8 error but pylint errors). I'm just making it consistent and 120 fits on my screen pretty much perfectly. 100 leads to a lot of unnecessary linebreaks when importing etc.

application/artifacts/artifacts_view.py
Askir: Patch and put should probably not do the same thing per default. I like having that explicitly. I will call the other method from the function though. So we don't have actual code duplication.

That way you can see that for example: PUT does actually just call Patch and when you want to have an actual PUT you'll have to reimplement it. Thoughts?

application/routes.py
Askir: Yes and move teams to users imo.

application/base.py
Askir: rip

application/artifacts/artifact_schema.py
Askir: I agree should we make an issue for API documentation with swagger?

application/artifacts/artifacts_view.py
Askir: actually... I'm not very consistent with that either. Let's talk about this and then make it consistent and put it into some sort of style guide :D

application/artifacts/artifacts_view.py
Askir: You can check users, teams and artifacts controller. Sometimes PATCH is actually just a PUT and sometimes it's vice versa. So I am calling the respective one.

application/base.py
Askir: done

Adrian-St (approved):

PR #84, opened 2019-03-14
Password page

luisebenkert:


Adrian-St (approved):

application/controllers/users_controller.py
```diff
@@ -44,6 +44,13 @@ def update(self, params, object_id):
         object_id = params.pop("id")
         try:
             user = User.find_by(id_=object_id)
+            if "password" in params:
+                password = params.pop("password")
+                old_password = params.pop("old_password", None)
+                success = user.update_password(password, old_password)
+                print(success)
```
Askir, line 51: There is a print here!!! :D

Askir: Removed it in the bigger refactoring though. Just wanted to send you an extra email so you feel judged!

PR #86, opened 2019-03-19
Google login

Adrian-St:
Adds Google login and ability to allow and revoke google drive access.


application/users/oauth_providers/google_oauth.py
```diff
@@ -0,0 +1,75 @@
...
+        google_id, _ = validate_google_id_token(id_token)
+        return google_id
+
+    def revoke_access(self, credentials):
+        credentials = Credentials(**credentials)
+
+        revoke = requests.post('https://accounts.google.com/o/oauth2/revoke',
+                               params={'token': credentials.token},
+                               headers={'content-type': 'application/x-www-form-urlencoded'})
+        print(getattr(revoke, 'status_code'))
```
Askir: WHAT IS THAT?

Askir (commented):
Generally nice functionality BUT:
I'd change the Datamodel to have only a "Googlexyz"-Node with the google specific fields (user_id, credentials) and connect that via a HAS_GOOGLE_OAUTH relationship.

Then the routes should work slightly differently as well.
- PUT `/users/:id/oauths/google` should create the google authentication
- (DELETE `/users/:id/oauths/google` should delete the authentication)
- PATCH `/users/:id/oauths/google` adds scopes
- DELETE `/users/:id/oauths/google/scopes` deletes them

application/authentications/authentications_view.py
```diff
@@ -23,10 +24,18 @@ class AuthenticationsView(MethodResource):
...
-        if not validate_user(user, params["password"]):
-            return {"error": "Bad username or password"}, 401
-        return self._build_login_response(user)
+        if "id_token" in params:
+            user = _get_user_from_google_token(params["id_token"])
+        else:
+            if "email_or_username" not in params:
+                return {"error": "email_or_username missing"}, 422
+            if "password" not in params:
+                return {"error": "password missing"}, 422
```
Askir: This does feel like replicating webargs logic. But you cannot make the parameters required as you have the google login on this route as well. You could check if there is a way to call the validation manually and still extract this logic though. Or build a custom validator that requires either the `id_token` or `username+passowrd`. Either way, I don't think this belongs here and should at least be method extracted.

application/users/__init__.py
```diff
@@ -2,7 +2,10 @@
 from flask import Blueprint
 
 from .users_view import UserView, UsersView
+from .oauth.google_oauth_view import GoogleOauthView, OAuthScopesView
 
 USERS = Blueprint('users', __name__)
 USERS.add_url_rule('', view_func=UsersView.as_view('usersview'))
 USERS.add_url_rule('/<id>', view_func=UserView.as_view('userview'))
+USERS.add_url_rule('/<id>/oauths/google', view_func=GoogleOauthView.as_view('googleoauthview'))
+USERS.add_url_rule('/<id>/oauths/google/scopes', view_func=OAuthScopesView.as_view('oauthscopesview'))
```
Askir: Can we rename the `OAuthScopesView` to `GoogleScopesView`? The scopes are not specific to oauth in general but to Google instead.

application/users/user.py
```diff
@@ -11,10 +12,21 @@ class User(StructuredNode, DefaultPropertyMixin, DefaultHelperMixin):  # pylint:
...
+    password = StringProperty()
 
     teams = RelationshipFrom('application.models.Team', 'HAS_MEMBER', cardinality=cardinality.ZeroOrMore)
     artifacts = RelationshipFrom('application.models.Artifact', 'CREATED_BY', cardinality=cardinality.ZeroOrMore)
+    google_rel = RelationshipTo('application.models.GoogleOAuth', 'HAS_GOOGLE_OAUTH', cardinality=cardinality.ZeroOrOne)
+
+    @property
+    def google(self):
+        """Returns GoogleOauth Node"""
+        return self.google_rel.single() # pylint:disable=no-member
```
Askir, line 24: This is really stupid, we should definitely make a PR on neomodel for that.

application/users/user.py
```diff
@@ -39,5 +51,5 @@ def update(self, **properties):
 
     def pre_save(self):
         super()
-        if self.does_not_exist():
+        if self.does_not_exist() and self.password:
```
Askir: I'd appreciate it if we could use the `has_password` from further up for consistency sake here.

application/users/user_schema.py
```diff
@@ -1,14 +1,16 @@
...
+from application.users.oauth.google_oauth_schema import GoogleOAuthSchema
 
 
 class UserSchema(BaseSchema):
     """Schema for importing and exporting Team objects"""
     id_ = fields.UUID(missing=None)
     username = fields.String()
     email = fields.String()
+    google = fields.Nested(GoogleOAuthSchema)
+    has_password = fields.Bool()
```
Askir, line 13: do we need to export has_password?

Askir (approved):
I like it now :D Just some styling and separation of concerns for view and validation. Thanks for baring with me!

application/users/user_schema.py
Adrian-St: yeah :( In the settings page a user can change the password, if he doesn't have a password because it's a google user, that shouldn't be rendered

PR #90, opened 2019-04-08
Invitation link

luisebenkert:


.gitignore
```diff
@@ -9,5 +9,8 @@ __pycache__
 uploads/
 htmlcov
 .venv
+localhost/
+teams.xlsx
+application/create_accounts.py
```
Askir: If you have local changes that you don't want to commit, you don't have to add it to the .gitignore which is meant to be used for everyone developing on the project. Instead you can use `.git/info/exclude` to specify your personal excludes.

application/teams/__init__.py
```diff
@@ -1,8 +1,10 @@
...
 from flask import Blueprint
 
-from .teams_view import TeamView, TeamsView
+from .teams_view import TeamView, TeamsView, MembersView, AllTeamsView
 
 TEAMS = Blueprint('teams', __name__)
 TEAMS.add_url_rule('', view_func=TeamsView.as_view('teamsview'))
 TEAMS.add_url_rule('/<id>', view_func=TeamView.as_view('teamview'))
+TEAMS.add_url_rule('/<id>/members', view_func=MembersView.as_view('membersview'))
+TEAMS.add_url_rule('/all', view_func=AllTeamsView.as_view('allteamsview'))
```
Askir: There should not be a public route to request all teams. Teams are specific to each user and at the current moment not publicly accessible, and `/all` is not RESTful (https://www.restapitutorial.com/lessons/restfulresourcenaming.html). See overall comment for eriks and my opinion on the matter.

application/teams/__init__.py
```diff
@@ -1,8 +1,10 @@
 """Provides Team functionality and routes"""
 from flask import Blueprint
 
-from .teams_view import TeamView, TeamsView
+from .teams_view import TeamView, TeamsView, MembersView, AllTeamsView
 
 TEAMS = Blueprint('teams', __name__)
 TEAMS.add_url_rule('', view_func=TeamsView.as_view('teamsview'))
 TEAMS.add_url_rule('/<id>', view_func=TeamView.as_view('teamview'))
+TEAMS.add_url_rule('/<id>/members', view_func=MembersView.as_view('membersview'))
```
Askir: Adding a member via POST is intuitively REST. But since you need to know the team_id to perform this action (which we don't have, we only have the join_key), it is not exactly what we need.

Askir (dismissed):
Erik and me talked about this a little bit. It is actually difficult to model an action like `join via key` in REST. REST is only defined for the default CRUD actions. Join however does not really fit Create, Read, Update or Delete very well.

We reflected upon a few options and decided on the following:
- Join via POST `/invites/:invite_key`

If this route is implemented. The all teams route is not needed anymore, and joining is handled via this instead of POSTING on `/members`.
___
We compared different approaches between APIs to see how they handle non CRUD route naming. For inspiration you can checkout:

- Stripe: https://stripe.com/docs/api
- Okta: https://developer.okta.com/docs/api/resources/apps/


We also tried out the invite links from discord and slack. And decided to use the discord approach to joining.
___

Here is also an interesting collection of ressources for [API design](https://techbeacon.com/app-dev-testing/guide-restful-api-design-35-must-reads). Just in case you want to dive a little deeper.

.gitignore
arne-z: 👍

application/invites/invites_view.py
```diff
@@ -0,0 +1,32 @@
...
+from flask_apispec.views import MethodResource
+from flask_jwt_extended import get_jwt_identity, jwt_required
+
+from application.teams.team import Team
+from application.teams.team_schema import TEAM_SCHEMA
+from application.users.user import User
+from .invites_validator import accept_invite_args
+
+
+class AcceptInvite(MethodResource):
```
erksch: Why do you not call it InvitesView?

application/invites/invites_view.py
```diff
@@ -0,0 +1,32 @@
...
+    """Controller for invite links."""
+
+    @jwt_required
+    @use_kwargs(accept_invite_args())
+    @marshal_with(TEAM_SCHEMA)
+    def post(self, **params):
+        """Logic for adding a single team member"""
+        print('JWT: ' + str(get_jwt_identity()))
```
erksch: Print

application/invites/invites_view.py
```diff
@@ -0,0 +1,32 @@
...
+        user = User.find_by(id_=get_jwt_identity())
+        invite_link = params['join_key']
+        team = Team.find_by(join_key=invite_link, force=False)
```
erksch: why the renaming?
just use 
`team = Team.find_by(join_key=params['join_key'], force=False)`

erksch: What does force do?

application/teams/team.py
```diff
@@ -1,14 +1,15 @@
...
 
 from application.model_mixins import DefaultPropertyMixin, DefaultHelperMixin
 from application.teams.team_schema import TeamSchema
 
-
 class Team(StructuredNode, DefaultPropertyMixin, DefaultHelperMixin):  # pylint:disable=abstract-method
     """The class that manages Teams"""
     schema = TeamSchema
     name = StringProperty(required=True)
+    join_key = StringProperty(required=False, default=uuid4)
```
erksch: Should the join key be a uuid? I think it could be a lot shorter?

specs/requests/invites_specs.py
```diff
@@ -0,0 +1,58 @@
...
+
+from expects import expect, have_key, equal, contain
+from mamba import description, before, after, it
+from neomodel import db
+
+from specs.factories.team_factory import TeamFactory
+from specs.factories.user_factory import UserFactory
+from specs.spec_helpers import Context
+
+sys.path.insert(0, 'specs')
```
erksch: What does this do?

specs/requests/invites_specs.py
```diff
@@ -0,0 +1,58 @@
...
+        db.cypher_query("MATCH (a) DETACH DELETE a")
+        pass
+
+    with description('/:id'):
+        with description('POST'):
+            with description('member not part of team'):
+                with before.each:
+                    self.team = TeamFactory.create_team(name='Blue')
+                    self.user = create_and_login_test_user(self.context.client())
+                    print('USER: ' + str(self.user))
```
erksch: No print.

specs/requests/invites_specs.py
```diff
@@ -0,0 +1,58 @@
...
+
+                with it('responds with 200'):
+                    expect(self.response.status_code).to(equal(200))
+
+                with it('responds with the correct team'):
+                    expect(self.response.json).to(have_key("name", "Blue"))
+
+                with it('includes added member'):
+                    expect(self.response.json).to(have_key("members"))
+                    expect([user['id'] for user in self.response.json['members']]).to(
```
erksch: Man what a line, maybe extract `user['id'] for user in self.response.json['members']` to an `member_ids` variable so it does not look so frightening.

application/invites/invites_view.py
erksch: But I get that then the post method would not really be clear as an acceptance of an invite.
@Askir what do you think?

specs/requests/invites_specs.py
```diff
@@ -0,0 +1,58 @@
...
+            with description('member already part of team'):
+                with before.each:
+                    self.team = TeamFactory.create_team(name='Blue')
+                    self.user = create_and_login_test_user(self.context.client())
+                    self.team.members.connect(self.user)
+                    self.response = self.context.client().post(
+                        "/invites/" + str(self.team.join_key))
+
+                with it('responds with 409'):
+                    expect(self.response.status_code).to(equal(409))
```
erksch, line 55: Nice testing in general :)

erksch (changes requested):

application/invites/invites_view.py
arne-z: 👍

application/invites/invites_view.py
arne-z: force=True causes the function to fail and raise an exception if the team is not found, whereas force=False will let it return none.
I am checking for team=None on the next line and returning a 404 if the team does not exist, so an exception is not needed at this point.

application/invites/invites_view.py
arne-z: I am renaming it to InvitesView for now, but I am not sure if that conveys the right idea.

specs/requests/invites_specs.py
arne-z: 👍

specs/requests/invites_specs.py
arne-z: @luisebenkert is to blame for that 😄

specs/requests/invites_specs.py
arne-z: I don't know, but It's at the beginning of all our test files...

specs/requests/invites_specs.py
arne-z: I needed a moment to understand that one as well. Implemented your change.

application/teams/team.py
arne-z: UUID was decided on after internal discussion.

erksch (approved):

application/teams/team.py
Askir: I'd prefer a shorter join link.
This could lead to potential collisions (but still very unlikely just like uuids). But it makes the links more human friendly and if we enable regeneration later on collision wont be a huge problem.
You can use this for that:
```
shortId = RandomStringUtils.randomAlphanumeric(8);
```
https://stackoverflow.com/questions/4267475/generating-8-character-only-uuids as reference

specs/requests/invites_specs.py
Askir: sys.path is the path python uses to import modules. Inserting 'specs' at 0 makes the specs folder the first one to look for for modules to import. Python should however automatically add the current file and folder to the top of the list, so I don't think this is needed. You can try removing it and see if everything still works.

If not it'd be interesting to see how and why that affects imports.

application/invites/invites_view.py
Askir: Imo `except NotFound` is more expressive and you don't have the extra force variable which apparently can be misunderstood.

application/invites/invites_view.py
Askir: InvitesView seems like a pretty general name that we'd use for all Restful resources.
However the plural view (InvitesView, UsersView, ArtifactsView) is always mounted under the colleciton path (/invites, /users, /artifacts). Since this handles the POST for a single Invite it should be `InviteView`.

Post on a single member of a resource is usually not defined for REST (https://www.restapitutorial.com/lessons/httpmethods.html). Some APIs use it as an alternative to PUT. But using it for a custom event is perfectly fine as long as it makes sense from an understanding point of view.

application/teams/team.py
Askir: We can make the property unique here and add an index (`unique_index=true`). That way team creation will fail if by any chance the generated key already exists.

specs/requests/invites_specs.py
Askir: Always important to question older decisions. Either you improve something or we learn something :P

Askir (commented):
Looks good over all. Just added a few minor remarks. I do prefer the shorter join key and I don't think it should be too much of a hassle to change it (schema, validator, model and tests factory would need changes). If there is a good reason for a uuid I'm open to that as well but seems not very userfriendly just incase you have to type it by hand at some point.

PR #91, opened 2019-04-10
Log search queries

henleo:


app.py
```diff
@@ -1,9 +1,14 @@
...
 import logging
+from application.custom_loggers import setup_file_logger
 
 from application import create_app
 from application.extensions import socketio
 
+setup_file_logger(name='query_logger', filename='search_query.log',
+                  format="%(asctime)s search query '%(search_query)s' "
+                  + "offset: %(offset)s limit: %(limit)s team: %(team)s user: %(user)s")
+
```
Adrian-St: The format is very tightly coupled with the search now. I would prefer something more generic, or even set the format when you log something.

application/artifacts/artifacts_view.py
```diff
@@ -125,11 +126,19 @@ class ArtifactsView(MethodResource):
...
     @marshal_with(ARTIFACTS_SCHEMA)
     def get(self, **params):
         """Logic for querying several artifacts"""
+        query_logger = logging.getLogger('query_logger')
+        query_logger.info("", extra={'search_query': str(params['search']),
+                                     'offset': str(params['offset']),
+                                     'limit': str(params['limit']),
+                                     'team': str(params['team_id']),
+                                     'user': str(get_jwt_identity())})
+
```
Adrian-St: Is there a reason why you don't log all params? If we refactor the search we have to change the logging too

app.py
Adrian-St: Also please put the logfile path in the env. We probably want it somewhere else in production.

application/custom_loggers.py
```diff
@@ -0,0 +1,16 @@
...
+def setup_file_logger(name, filename, format='%(asctime)s %(message)s'):
+    """Configures a custom logger that writes to a file and returns it"""
+    logger = logging.getLogger(name)
+    logger.setLevel(logging.DEBUG)
+    handler = logging.FileHandler(filename)
+    formatter = logging.Formatter(format)
+    handler.setFormatter(formatter)
+    handler.setLevel(logging.DEBUG)
+    logger.addHandler(handler)
+    return logger
```
Adrian-St, line 17: If you have the logger in app.py I would put that file outside of application.
Since the logger is really specific I think it would be okay to move everything into application too.

Adrian-St (changes requested):
Either make the logger more generic or move the logger setup into /application/aritfacts/__init__.py please.

Adrian-St (approved):

PR #92, opened 2019-04-10
Add password reset

Adrian-St:
Not sure about the routing.
I think I would like it better if it was at /users/reset but that is not that restful.


.env.example
```diff
@@ -6,6 +6,9 @@ CLOUD_VISION_API_URL=https://vision.googleapis.com/v1/images:annotate
 CLOUD_VISION_API_KEY=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 JWT_SECRET_KEY=secret_key
 SECURE_COOKIES=False
-CLIENT_IDS_PATH=clientIds.json
-CLIENT_SECRET_PATH=credentials.json
+CLIENT_IDS_PATH=clientIds.json # Path to json in the form of [ "clientId1", "clientId2", ... ]
+CLIENT_SECRET_PATH=credentials.json  # Path to credentials json. Downloadable from google developer console
```
Askir, line 10: Thanks for these. Very useful :D

application/password_resets/password_resets_validator.py
```diff
@@ -0,0 +1,19 @@
+"""Defines validators for password reset requests"""
+
+from webargs import fields
+
+
+def create_args():
```
Askir: These aren't really create requests right?

application/password_resets/password_resets_validator.py
```diff
@@ -0,0 +1,19 @@
...
+    """Defines and validates params for create"""
+    return {
+        "email_or_username": fields.String(required=True),
+        "base_url": fields.URL(required=True)
+    }
+
+
+def update_args():
```
Askir: And neither is this an actual update request.

application/password_resets/password_resets_view.py
```diff
@@ -0,0 +1,64 @@
...
+    @use_kwargs(password_resets_validator.create_args())
+    def post(self, **params):
+        """Logic for sending a password reset link"""
+        user = User.find_by_email_or_username(params["email_or_username"])
+        if not user:
+            abort(404, "Could not find a user with that email address")
+        token = password_reset_jwt_manager.encode_reset_token(user)
+        reset_url = set_query_parameter(params["base_url"], "reset_token", token)
+        send_reset_password_email(reset_url, user)
+        return no_content()
```
Askir, line 29: Should this really be a no content response? How about 201 (created reset), 200(just general okay) or 202 (Accepted request ready for processing).

Askir (commented):
Looks nice but I don't quite understand all the jwt reset magic. Would you mind explaining what's actually happening a bit?

application/password_resets/password_resets_view.py
erksch: No Content is mostly just used for a Delete request. 201 seems suitable here.

Adrian-St: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201
I am not really creating a new resource (only if you think of the email as an resource) and I don't return anything. 201 should return something. Maybe a 200 would be okay here, but nothing says that 204 is reserved for delete

Askir: discord uses 204 here as well. So I'm fine with it :D

PR #93, opened 2019-04-11
Proposed changes to linting setup.

arne-z:


Askir (changes requested):
generally nice idea, but I don't like that so many unnecessary linebreaks are added (is the line length configurable to 100 or 120 isntead of probably 80 right now?). And I would also prefer single quoted Strings.

arne-z:
@Askir 
Line length is the default that Black uses atm, which is 88. They have some [reasoning](https://github.com/ambv/black#line-length) for that, but I would not mind setting the line length to a higher value. We were previously using max-line-length=120 in our Pylintrc and I see no problem continuing that.

On the matter of double quotes vs. single quotes I actually really like the approach and decisions by black and would be strongly in favor of following their [advice](https://github.com/ambv/black#strings).

PR #98, opened 2019-04-17
Refactor Artifact creation.

Askir:
I''ve extracted all the creation logic into a seperate file. Which will hopefully enable me to reuse that logic when downloading files from google drive, without sending actual POST requests to /images.

I also renamed the ArtifactBuilder to ArtifactConnector so it doesn't get confused with ArtifactCreator. If there are any other naming suggestions. I'm very open to them. Seems at least slightly confusing to me as well still.

### DIFF
--- a/application/artifacts/artifact_connector.py
+++ b/application/artifacts/artifact_connector.py
@@ -1,18 +1,17 @@
-"""Abstracting Sync of Neo and Elastic"""
-
 from application.artifacts.artifact import Artifact
+from application.artifacts.property_builder import PropertyBuilder
 from application.artifacts.tags.tag import Tag
 from application.teams.team import Team
 from application.users.user import User
-from application.artifacts.property_builder import PropertyBuilder
 
 
-class ArtifactBuilder:
-    """This class helps with building Artifacts"""
+class ArtifactConnector:
+    """This class helps with building the database connections
+    for Artifact model objects."""
 
     @classmethod
     def for_artifact(cls, neo):
-        """Create an ArtifactBuilder for a specific Artifact"""
+        """Create an ArtifactConnector for a specific Artifact"""
         builder = cls()
         builder.neo = neo
         return builder
@@ -21,7 +20,7 @@ class ArtifactBuilder:
         self.neo = None
 
     def build_with(self, **properties):
-        """builds self.neo and saves it"""
+        """builds a new Artifact with correct connections and saves it"""
         props = PropertyBuilder(**properties)
         neo_properties = props.node_properties
         self.neo = Artifact(**neo_properties).save()
@@ -37,7 +36,7 @@ class ArtifactBuilder:
             self._connect_relation(self.neo.user, User.find(props["user_id"]))
 
     def save(self):
-        """Save both models"""
+        """Saves the referenced artifact model"""
         self.neo.save()
 
     def _add_tags_to_neo(self, new_tags, tag_relation, override=False):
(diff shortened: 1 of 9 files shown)


application/artifacts/artifact_connector.py
```diff
@@ -66,7 +64,7 @@ def update_with(self, override_tags=True, **properties):
     @classmethod
     def find(cls, uid, force=True):  # pylint: disable= invalid-name
```
arne-z: `  # pylint: disable= invalid-name`
should no longer be necessary.

application/artifacts/artifact_creation.py
```diff
@@ -0,0 +1,110 @@
...
+import werkzeug
+from PIL import Image
+from flask import current_app
+
+from application.artifacts.artifact_connector import ArtifactConnector
+
+from application.artifacts.image_recognition import ImageRecognizer
+
+
+class ImageResizer:  # pylint:disable=too-few-public-methods
```
arne-z: `# pylint:disable=too-few-public-methods`
should not be necessary.

application/artifacts/artifact_creation.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def save_sizes(self):
+        """Saves resized image for all defined widths"""
+        for width in self.widths:
+            height = self._calculate_height(width)
+            resized_image = self.image.resize((width, height), Image.ANTIALIAS)
+            resized_image.save(self._generate_file_path(width))
+            resized_image.close()
+
+    def _generate_file_path(self, width):
```
arne-z, line 31: Could add docstring here just for consistency.

application/artifacts/artifact_connector.py
```diff
@@ -1,18 +1,16 @@
...
+from application.artifacts.property_builder import PropertyBuilder
 from application.artifacts.tags.tag import Tag
 from application.teams.team import Team
 from application.users.user import User
-from application.artifacts.property_builder import PropertyBuilder
 
 
-class ArtifactBuilder:
+class ArtifactConnector:
     """This class helps with building Artifacts"""
```
arne-z: Could make docstring more specific (explain why it's a connector).

application/artifacts/artifact_creation.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class ArtifactCreator:
+    def __init__(self, file, owner_id, team_id, **optional):
+        self.file = file
+        self.owner_id = owner_id
+        self.team_id = team_id
+        self.additional_params = optional
+
+    def create_artifact(self):
```
arne-z, line 94: Docstring?

specs/requests/artifacts_spec.py
```diff
@@ -38,7 +38,7 @@ def clear_upload_dir():
         self.context.client().login(self.user)
 
     with after.each:
-        clear_upload_dir()
+        # clear_upload_dir()
```
arne-z: This can't be right?

arne-z (changes requested):

application/artifacts/artifact_creation.py
Askir: I wont add a docstring to private methods (except if it isn't clear what the are doing by method name). But I added a class comment and one to the public method.

specs/requests/artifacts_spec.py
Askir: nope, good catch! :D

Adrian-St (approved):
I like it :)

application/artifacts/artifact_creation.py
arne-z: 👍

arne-z (approved):

PR #101, opened 2019-04-25
Fixes socketio

Adrian-St:
Change respond_with to marshal_data and send search string when searching with socketio


application/artifacts/artifacts_view.py
```diff
@@ -29,16 +29,27 @@
...
+    return data
 
 
 def _search_artifacts(params):
     search_args = params.get("search")
     if search_args is not None:
         params["synonyms"] = SynonymGenerator(search_args).get_synonyms()
         elastic_artifacts = ElasticSearcher.build_artifact_searcher(params).search()
-
+        print(elastic_artifacts)
```
erksch: lolol

erksch (changes requested):

